PR #18, opened 2020-12-30
Make Datagram-Flow-Id a list

### LINKED ISSUE #9: Field:Flow ID cardinality
It's not really clear whether there are any expectations in relation to cardinality of flow identifiers and the field.  Can two fields refer to the same flow ID?  Can the same request reference multiple flow IDs?

### LINKED ISSUE #11: SF example parameter is confusing
The document currently uses the following example:

```
     Datagram-Flow-Id = 42; alternate=44
```
Motivating this with an example of how to express multiple flow identifiers.

I think that this example is potentially harmful.  It implies a lot about a potential extension mechanism as part of an example, but does not commit to properly defining an extension.  For examples like this I would much prefer to see something that is very clearly bogus.  Otherwise, people might implement that to the point that they even achieve interoperability of a sort.

For instance, the following requires a lot more guessing about how to interpret it, to the point that I would guess that pseudo-interoperability would be hard to arrive at by accident:

```
     Datagram-Flow-Id = 42; camels-per-orthodoxy=17.4
```

See also #9.

### LINKED ISSUE #12: Many reasons why a message with Datagram-flow-Id could be delayed
Section 6 makes the keen observation that:

```
   Since the QUIC STREAM frame that contains the "Datagram-Flow-Id"
   header could be lost or reordered, it is possible that an endpoint
   will receive an HTTP/3 datagram with a flow identifier that it does
   not know as it has not yet received the corresponding "Datagram-Flow-
   Id" header.  Endpoints MUST NOT treat that as an error; they MUST
   either silently discard the datagram or buffer it until they receive
   the "Datagram-Flow-Id" header.
```

This is accurate but, when thinking about HTTP/3 APIs, does not cover all possible reasons that might cause a request message to be processed after the corresponding DATAGRAM frame is available. For instance, the message might be blocked by QPACK, or the implementation might present DATAGRAMS before stream headers. 

We could boil the ocean with describing all possible reasons, so I think it might be better to shuffle the text to focus on describe the symptom (DATAGRAMS before HTTP messages) and the treatment. We can then give non-exhaustive example(s) as already done.

### LINKED ISSUE #13: Running out of Flow IDs
https://tools.ietf.org/html/draft-schinazi-masque-h3-datagram-01#section-6 states:

```
   Note that integer structured fields can only encode values up to
   10^15-1, therefore the maximum possible value of the "Datagram-Flow-
   Id" header is lower then the theoretical maximum value of a flow
   identifier which is 2^62-1 due to the QUIC variable length integer
   encoding.  If the flow identifier allocation service of an endpoint
   runs out of values lower than 10^15-1, the endpoint MUST treat is as
   a connection error of type H3_ID_ERROR.
```

The text is a little ambiguous. I presume the intention is that the endpoint that is allocating flows should do what HTTP/3 defines as [Immediate Application Closure](https://tools.ietf.org/html/draft-ietf-quic-http-32#section-5.3)? The doesn't cover the case where an endpoint receives a Datagram Flow ID greater than 10^15-1 (either in a request or a DATAGRAM).

I think it might help to separate these cases. Receiving a value of 10^15-1 is always a connection error of type H3_ID_ERROR. 
But I'm not so sure about the sender case, if you really are flow ID exhausted, perhaps it is nicer to just spin up a new connection and leave the current one as is. We might recommend that endpoints initiate an HTTP/3 graceful close followed by an application close - but reusing H3_ID_ERROR for that case seems a bit overkill.

DavidSchinazi:
This PR switches the header to a list on MT's suggestion to make the semantics clearer.

Closes #9
Closes #11
Closes #12
Closes #13

### DIFF
--- a/draft-schinazi-masque-h3-datagram.md
+++ b/draft-schinazi-masque-h3-datagram.md
@@ -165,51 +165,61 @@ with error H3_SETTINGS_ERROR.
 # Datagram-Flow-Id Header Field Definition {#header}
 
 "Datagram-Flow-Id" is a Item Structured
-Field {{!STRUCT-FIELD=I-D.ietf-httpbis-header-structure}}. Its value MUST be an
-Integer. Its ABNF is:
+Field {{!STRUCT-FIELD=I-D.ietf-httpbis-header-structure}}. Its value MUST be a
+List of Integers. Its ABNF is:
 
 ~~~
-  Datagram-Flow-Id = sf-integer
+  Datagram-Flow-Id = sf-list
 ~~~
 
-The "Datagram-Flow-Id" header field is used to associate a datagram flow
-identifier with an HTTP message. For example, the definition of an HTTP method
-could instruct the client to use its flow identifier allocation service to
-allocate a new flow identifier, and then the client will add the
-"Datagram-Flow-Id" header field to its request to communicate that value to the
-server. For example, the resulting header field could look like:
+The "Datagram-Flow-Id" header field is used to associate one or more datagram
+flow identifiers with an HTTP message. As a simple example using a single
+identifier, the definition of an HTTP method could instruct the client to use
+its flow identifier allocation service to allocate a new flow identifier, and
+then the client will add the "Datagram-Flow-Id" header field to its request to
+communicate that value to the server. In this example, the resulting header
+field could look like:
 
 ~~~
   Datagram-Flow-Id = 2
 ~~~
 
-Definitions of HTTP features that use the "Datagram-Flow-Id" header field MAY
-define their own parameters (parameters are defined in Section 3.1.2 of
-{{STRUCT-FIELD}}). For example, an HTTP method that wishes to use two datagram
-flow identifiers for the lifetime of its request stream could encode the second
-flow identifier as a parameter, which could look like this:
+One element in the list is allowed to be unnamed, but all but one elements
+MUST carry a name. The name is conveyed using a true boolean parameter
+(parameters are defined in Section 3.1.2 of {{STRUCT-FIELD}}) which MUST
+bew the first parameter of that element. The ordering of the list does not
+carry any semantics. For example, an HTTP method that wishes to use four
+datagram flow identifiers for the lifetime of its request stream could look
+like this:
 
 ~~~
-  Datagram-Flow-Id = 42; alternate=44
+  Datagram-Flow-Id = 42, 44; ecn-ect0, 48; ecn-ce, 46; ecn-ect1
 ~~~
 
-The "Datagram-Flow-Id" header field MUST NOT be present more than once on a given
-HTTP message; any HTTP message containing more than one "Datagram-Flow-Id"
-header field is malformed.
+In this example, 42 is the unnamed flow identifier, 44 represents "ecn-ect0",
+46 represents "ecn-ect1" and 48 represents "ecn-ce".
 
-Since the QUIC STREAM frame that contains the "Datagram-Flow-Id" header field could
-be lost or reordered, it is possible that an endpoint will receive an HTTP/3
-datagram with a flow identifier that it does not know as it has not yet
-received the corresponding "Datagram-Flow-Id" header field. Endpoints MUST NOT treat
-that as an error; they MUST either silently discard the datagram or buffer it
-until they receive the "Datagram-Flow-Id" header field.
+Even if a sender attempts to communicate the meaning of a flow identifier
+before it uses it in an HTTP/3 datagram, it is possible that its peer will
+receive an HTTP/3 datagram with a flow identifier that it does not know as it
+has not yet received the corresponding "Datagram-Flow-Id" header field. (For
+example, this could happen if the QUIC STREAM frame that contains the
+"Datagram-Flow-Id" header field is reordered and arrives afer the DATAGRAM
+frame.) Endpoints MUST NOT treat that scenario as an error; they MUST either
+silently discard the datagram or buffer it until they receive the
+"Datagram-Flow-Id" header field.
+
+Distinct HTTP requests MAY refer to the same flow identifier in their
+respective "Datagram-Flow-Id" header fields.
 
 Note that integer structured fields can only encode values up to 10^15-1,
 therefore the maximum possible value of the "Datagram-Flow-Id" header field is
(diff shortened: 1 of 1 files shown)


draft-schinazi-masque-h3-datagram.md
```diff
@@ -165,51 +165,61 @@ with error H3_SETTINGS_ERROR.
...
 
 Note that integer structured fields can only encode values up to 10^15-1,
 therefore the maximum possible value of the "Datagram-Flow-Id" header field is
 lower then the theoretical maximum value of a flow identifier which is 2^62-1
 due to the QUIC variable length integer encoding. If the flow identifier
 allocation service of an endpoint runs out of values lower than 10^15-1, the
-endpoint MUST treat is as a connection error of type H3_ID_ERROR.
+endpoint MUST fail the flow identifier allocation. An HTTP message that
+carries a "Datagram-Flow-Id" header field with a flow identifier value above
+10^15-1 is malformed.
```
LPardue, line 222: This is good. Is there something in Semantics we can cite to explain how malformed messages are handled?

draft-schinazi-masque-h3-datagram.md
```diff
@@ -165,51 +165,61 @@ with error H3_SETTINGS_ERROR.
...
+MUST carry a name. The name is conveyed using a true boolean parameter
+(parameters are defined in Section 3.1.2 of {{STRUCT-FIELD}}) which MUST
+bew the first parameter of that element. The ordering of the list does not
+carry any semantics. For example, an HTTP method that wishes to use four
+datagram flow identifiers for the lifetime of its request stream could look
+like this:
 
 ~~~
-  Datagram-Flow-Id = 42; alternate=44
+  Datagram-Flow-Id = 42, 44; ecn-ect0, 48; ecn-ce, 46; ecn-ect1
```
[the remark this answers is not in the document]
DavidSchinazi: I ended up splitting the example in two: first the logical ordering then your example which nicely shows off that the unnamed can be in any position

draft-schinazi-masque-h3-datagram.md
```diff
@@ -165,51 +165,61 @@ with error H3_SETTINGS_ERROR.
 # Datagram-Flow-Id Header Field Definition {#header}
 
 "Datagram-Flow-Id" is a Item Structured
-Field {{!STRUCT-FIELD=I-D.ietf-httpbis-header-structure}}. Its value MUST be an
-Integer. Its ABNF is:
+Field {{!STRUCT-FIELD=I-D.ietf-httpbis-header-structure}}. Its value MUST be a
+List of Integers. Its ABNF is:
```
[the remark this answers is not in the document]
DavidSchinazi: Good call, done

draft-schinazi-masque-h3-datagram.md
DavidSchinazi: Unfortunately I didn't find anything in Semantics, so I referenced 7540. It's not perfect but it'll do.

draft-schinazi-masque-h3-datagram.md
```diff
@@ -165,51 +165,61 @@ with error H3_SETTINGS_ERROR.
...
 
-Definitions of HTTP features that use the "Datagram-Flow-Id" header field MAY
-define their own parameters (parameters are defined in Section 3.1.2 of
-{{STRUCT-FIELD}}). For example, an HTTP method that wishes to use two datagram
-flow identifiers for the lifetime of its request stream could encode the second
-flow identifier as a parameter, which could look like this:
+One element in the list is allowed to be unnamed, but all but one elements
+MUST carry a name. The name is conveyed using a true boolean parameter
+(parameters are defined in Section 3.1.2 of {{STRUCT-FIELD}}) which MUST
+bew the first parameter of that element. The ordering of the list does not
```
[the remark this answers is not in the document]
DavidSchinazi: Good call, I tried to clarify the text

draft-schinazi-masque-h3-datagram.md
```diff
@@ -165,51 +165,71 @@ with error H3_SETTINGS_ERROR.
 # Datagram-Flow-Id Header Field Definition {#header}
 
 "Datagram-Flow-Id" is a Item Structured
-Field {{!STRUCT-FIELD=I-D.ietf-httpbis-header-structure}}. Its value MUST be an
-Integer. Its ABNF is:
+Field {{!STRUCT-FIELD=I-D.ietf-httpbis-header-structure}}. Its value MUST be a
+List Structured Field, whose members MUST all be Items of type Integer. Its
```
LPardue, line 169: I think its simpler to write just

"Datagram-Flow-Id" is a List Structured Field {{!STRUCT-FIELD=I-D.ietf-httpbis-header-structure}}, whose members MUST all be Items of type Integer. Its ABNF is:"

(GitHub prevents me from making this as an actual direct suggestion)

draft-schinazi-masque-h3-datagram.md
```diff
@@ -165,51 +165,71 @@ with error H3_SETTINGS_ERROR.
...
+receive an HTTP/3 datagram with a flow identifier that it does not know as it
+has not yet received the corresponding "Datagram-Flow-Id" header field. (For
+example, this could happen if the QUIC STREAM frame that contains the
+"Datagram-Flow-Id" header field is reordered and arrives afer the DATAGRAM
+frame.) Endpoints MUST NOT treat that scenario as an error; they MUST either
+silently discard the datagram or buffer it until they receive the
+"Datagram-Flow-Id" header field.
+
+Distinct HTTP requests MAY refer to the same flow identifier in their
+respective "Datagram-Flow-Id" header fields.
```
LPardue, line 223: Do flow identifiers have to use the same name (or stay unamed) across requests? I think it would help to be strict here.

draft-schinazi-masque-h3-datagram.md
DavidSchinazi: Oh I should have read my work more carefully, you're absolutely right. Done.

draft-schinazi-masque-h3-datagram.md
DavidSchinazi: Discussed offline, the draft needs editorial work to clarify how to use names and their conceptual meaning, but we'll address that in a separate PR.

Issue #27, opened 2021-02-21
Make it explicit that Flow IDs are non-negative


LPardue:
The draft doesn't do this. It relies on the QUIC varint encoding for non-negativity in DATAGRAM frames but `sf-integer` could contain a negative value. Let's be clear in the spec.

DavidSchinazi:
Definitely agree, could you write a PR please?

LPardue:
I will.

LPardue:
fixed in https://github.com/ietf-wg-masque/draft-ietf-masque-h3-datagram/pull/29

PR #29, opened 2021-03-05
Split flow identifier term to flows and IDs

LPardue:
This somewhat of an experiment. While trying to address #27 I got distracted by the document's use of the "flow identifier" term.

The document likes to compare "flow identifiers" with QUIC streams, which is fine. But QUIC makes the distinction between streams and stream IDs; the ID is a property of the stream but it isn't the only one, and sometimes it helps to talk about the stream in other contexts.

This change tries to channel a similar editorial style. It replaces:

* "flow identifiers" with "flows" when talking about a logical entitiy
* "flow identifiers" with "flow IDs" when talking about the wire format, encoding, examples etc.

This could be a change too far but I figured it would be a good discussion starter.

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -32,11 +32,12 @@ author:
 The QUIC DATAGRAM extension provides application protocols running over QUIC
 with a mechanism to send unreliable data while leveraging the security and
 congestion-control properties of QUIC. However, QUIC DATAGRAM frames do not
-provide a means to demultiplex application contexts. This document defines how
+provide a means to demultiplex application contexts. This document describes how
 to use QUIC DATAGRAM frames when the application protocol running over QUIC is
-HTTP/3 by adding an identifier at the start of the frame payload. This allows
-HTTP messages to convey related information using unreliable DATAGRAM frames,
-ensuring those frames are properly associated with an HTTP message.
+HTTP/3. It defines logical flows identified by a non-negative integer that are
+present at the start of the DATAGRAM frame payload. Flows are associated with
+HTTP messages using the Datagram-Flow-Id header field, allowing endpoints to
+match unreliable DATAGRAMS frames to the HTTP messages that they are related to.
 
 Discussion of this work is encouraged to happen on the MASQUE IETF mailing list
 ([masque@ietf.org](mailto:masque@ietf.org)) or on the GitHub repository which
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-h3-datagram.md
```diff
@@ -76,46 +78,48 @@ document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
...
+# Flows
 
-Flow identifiers represent bidirectional flows of datagrams within a single QUIC
-connection. These are conceptually similar to streams in the sense that they
-allow multiplexing of application data.  Flows lack any of the ordering
-or reliability guarantees of streams.
+Flows are bidirectional exchanges of datagrams within a single QUIC connection.
+These are conceptually similar to streams in the sense that they allow
+multiplexing of application data. Flows are identified within a connection by a
+numeric value, referred to as the flow ID.  A flow ID is a 62-bit integer (0
```
DavidSchinazi, line 86:
```suggestion
numeric value, referred to as the flow identifier (or flow ID).  A flow ID is a 62-bit integer (0
```

draft-ietf-masque-h3-datagram.md
```diff
@@ -76,46 +78,48 @@ document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
 when, and only when, they appear in all capitals, as shown here.
 
 
-# Flow Identifiers {#flow-id}
+# Flows
```
DavidSchinazi, line 81: Do we want to call them "flows" or "datagram flows"? The second might be slightly clearer, albeit more wordy.

draft-ietf-masque-h3-datagram.md
```diff
@@ -76,46 +78,48 @@ document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
...
+Flows lack any of the ordering or reliability guarantees of streams. Beyond
+this, a sender SHOULD ensure that DATAGRAM frames within a single flow are
+transmitted in order relative to one another. If multiple DATAGRAM frames can be
+packed into a single QUIC packet, the sender SHOULD group them by flow to
+promote fate-sharing within a specific flow and improve the ability to process
+batches of datagram messages efficiently on the receiver.
 
 
-# Flow Identifier Allocation {#flow-id-alloc}
+# Flow Allocation {#flow-id-alloc}
```
DavidSchinazi, line 97: Technically, you don't allocate a flow, you allocate a flow ID - which is part of what you need to do when creating a new flow. So while I like the concept of a flow in general, I think the allocation service should deal in identifiers. What do you think?

DavidSchinazi (commented):
Overall I think this is a good improvement, thanks!

draft-ietf-masque-h3-datagram.md
LPardue: This text is liberally copied from QUIC transport stream definition. Im happy to diverge but just wanted to point that out.

draft-ietf-masque-h3-datagram.md
LPardue: I thought the same initially, I could go either way because the wordings doesn't bother me.

draft-ietf-masque-h3-datagram.md
LPardue: I think this is probably the most controversial part, so I'm glad we can agree it needs discussion.

From an implementation perspective I do allocate for flows but that is an application-context dependent property (CONNECT-UDP would setup a flow<>UDP socket mapping for instance). But you're right that this text is specifically about IDs and I think that ignoring application details in this document is probably fine.

Let me sleep on things and circle back.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: I can live with flow ID for consistency

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Maybe we go with datagram flow then since it's slightly clearer and matches the header name?

draft-ietf-masque-h3-datagram.md
LPardue: I've changed this to `Flow ID allocation service` throughout

draft-ietf-masque-h3-datagram.md
LPardue: agreed, done

DavidSchinazi (approved):
Thanks!

PR #37, opened 2021-03-24
Experiment with HTTP/3 frames

DavidSchinazi:
This PR is a thought experiment aimed at seeing whether using HTTP/3 frames for H3DGRAM and CONNECT-UDP makes sense. We're not planning on merging this as-is.

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -195,7 +195,7 @@ REGISTER_DATAGRAM_FLOW_ID Frame {
   Type (i) = TBD,
   Length (i),
   Flow ID (i),
-  Encoded Field Section (..),
+  Extension String (..),
 }
 ~~~
 {: #register-frame-format title="REGISTER_DATAGRAM_FLOW_ID HTTP/3 Frame Format"}
@@ -207,10 +207,19 @@ Flow ID:
 
 : The flow ID to associate with the QUIC stream that carried this frame.
 
-Encoded Field Section:
+Extension String:
 
-: QPACK-encoded fields. This allows extensions to optionally attach headers to
-their flow IDs.
+: A string of comma-separated key-value pairs to enable extensibility.
+
+The ABNF for the Extension String field is as follows (using syntax from
+{{Section 3.2.6 of RFC7230}}):
+
+~~~
+  extension-string = [ ext-member *( "," ext-member ) ]
+  ext-member       = ext-member-key "=" ext-member-value
+  ext-member-key   = token
+  ext-member-value = token
+~~~
 
 Note that these registrations are unilateral and unidirectional: the sender of
 the frame unilateraly defines the semantics it will apply to the datagrams it
@@ -294,6 +303,32 @@ This document will request IANA to register the following entry in the
   | H3_DATAGRAM  | 0x276 | This Document |    0    |
   +--------------+-------+---------------+---------+
 ~~~
+
+
+## Flow Extension Keys {#iana-keys}
+
+This document will request IANA to create an "HTTP Datagram Flow
+Extension Keys" registry. Registrations in this registry MUST
+include the following fields:
+
+Key:
+
+: The key (see {{register-frame}}). Keys MUST be valid tokens as defined in
+{{Section 3.2.6 of RFC7230}}.
+
+Description:
+
+: A brief description of the key semantics, which MAY be a summary if a
+specification reference is provided.
+
+Reference:
+
+: An optional reference to a specification for the parameter. This field MAY be
+empty.
+
+Registrations follow the "First Come First Served" policy (see Section 4.4 of
+{{!IANA-POLICY=RFC8126}}) where two registrations MUST NOT have the same Key.
+This registry is initially empty.
 
 
 --- back
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -36,7 +36,7 @@ provide a means to demultiplex application contexts. This document describes how
 to use QUIC DATAGRAM frames when the application protocol running over QUIC is
 HTTP/3. It defines logical flows identified by a non-negative integer that are
 present at the start of the DATAGRAM frame payload. Flows are associated with
-HTTP messages using the Datagram-Flow-Id header field, allowing endpoints to
+QUIC streams using the REGISTER_DATAGRAM_FLOW_ID HTTP/3 frame, allowing endpoints to
 match unreliable DATAGRAMS frames to the HTTP messages that they are related to.
 
 Discussion of this work is encouraged to happen on the MASQUE IETF mailing list
@@ -57,8 +57,8 @@ provide a means to demultiplex application contexts. This document describes how
 to use QUIC DATAGRAM frames when the application protocol running over QUIC is
 HTTP/3 {{!H3=I-D.ietf-quic-http}}. It defines logical flows identified by a
 non-negative integer that are present at the start of the DATAGRAM frame
-payload. Flows are associated with HTTP messages using the Datagram-Flow-Id
-header field, allowing endpoints to match unreliable DATAGRAMS frames to the
+payload. Flows are associated with QUIC streams using the REGISTER_DATAGRAM_FLOW_ID
+HTTP/3 frame, allowing endpoints to match unreliable DATAGRAMS frames to the
 HTTP messages that they are related to.
 
 This design mimics the use of Stream Types in HTTP/3, which provide a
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-h3-datagram.md
```diff
@@ -181,74 +181,82 @@ all cases, the maximum permitted value of the H3_DATAGRAM SETTINGS parameter is
...
-that wishes to use four datagram flows for the lifetime of its
-request stream could look like this:
+: The flow ID to associate with the QUIC stream that carried this frame.
 
-~~~
-  Datagram-Flow-Id = 42, 44; ecn-ect0, 46; ecn-ect1, 48; ecn-ce
-~~~
+Encoded Field Section:
+
+: QPACK-encoded fields. This allows extensions to optionally attach headers to
```
LPardue, line 212: I think this needs some more explanation; what use cases motivate this extensibility?

Unless there is a good reason to use QPACK I wouldn't. QPACK is for HTTP message metadata and these are clearly not those. I can predict this to cause all sorts of oddites/attacks when people try to reconstruct this field section and accidentally combine it with one reconstructed from HEADERS - or just the simple case of blocking/poisoning the QPACK dynamic table.

Could we satisfy the extensibility requirements with a simple value-only field of `sf-list`?

draft-ietf-masque-h3-datagram.md
```diff
@@ -181,74 +181,82 @@ all cases, the maximum permitted value of the H3_DATAGRAM SETTINGS parameter is
...
+their flow IDs.
+
+Note that these registrations are unilateral and unidirectional: the sender of
+the frame unilateraly defines the semantics it will apply to the datagrams it
+receives. If a mechanism using this feature wants to send datagrams of a given
+flow ID in both directions, this frame will need to be exchanged in both
+directions.
```
LPardue, line 219: I understand the spirit of the text but it seems slightly obtuse. It took a few reads and a cross-ref to CONNECT-UDP to help me. Maybe there is a some rewording to be done if we go in this direction.

LPardue:
General comment: Something you don't mention explicitly is whether it is permitted to send multiple REGISTER_DATAGRAM_FLOW_ID frames with different flow IDs on the same stream.

draft-ietf-masque-h3-datagram.md
ianswett: Agreed, it'd be good to understand what our extensibility requirements are here.  I can imagine a fairly wide range.

ianswett (commented):
Thanks David, some initial thoughts.

draft-ietf-masque-h3-datagram.md
```diff
@@ -181,74 +181,82 @@ all cases, the maximum permitted value of the H3_DATAGRAM SETTINGS parameter is
...
 
-"Datagram-Flow-Id" is a List Structured
-Field {{!STRUCT-FIELD=I-D.ietf-httpbis-header-structure}}, whose members MUST
-all be Items of type Integer. Integers MUST be non-negative. Its ABNF is:
+The REGISTER_DATAGRAM_FLOW_ID frame (type=TBD) is used to register a flow ID
+with the QUIC stream that it is sent on. By sending this frame, the sender
+indicates that it will process any received datagram with the corresponding
+flow ID using semantics specific to this frame. For example, if this is sent on
+a client-initiated bidirectional stream that has carried an HTTP request, this
+flow ID will now be interpreted as governed by the method of that request.
```
afrind, line 191: Does method always govern how DATAGRAMs are interpreted on a request/response stream?  Couldn't there be other mechanisms for specifying that?

draft-ietf-masque-h3-datagram.md
afrind: I agree we should be looking at what the extensibility requirements are, and providing the appropriate degree of freedom here.

I disagree that QPACK should only be used for HTTP message metadata and disallowed in extensions.  Why are we worried that this field section would be combined incorrectly with the header field section, but we're not worried about that for trailers?  Similar question for blocking/poisoning the QPACK table?

draft-ietf-masque-h3-datagram.md
afrind: I think the text should also address collisions or re-use of flow IDs.  What happens if the same flow ID is registered to more than one stream?  What is the lifetime of a flow ID -- does it auto-retire at the end of a stream, or can it outlive the FIN?  If the latter, how does the receiver of this frame know that the flow ID is done and state can be removed?

draft-ietf-masque-h3-datagram.md
```diff
@@ -181,74 +181,82 @@ all cases, the maximum permitted value of the H3_DATAGRAM SETTINGS parameter is
...
+receives. If a mechanism using this feature wants to send datagrams of a given
+flow ID in both directions, this frame will need to be exchanged in both
+directions.
 
-In this example, 42 is the unnamed flow, 44 represents the name
-"ecn-ect0", 46 represents "ecn-ect1", and 48 represents "ecn-ce". Note that,
-since the list ordering does not carry semantics, this example can be
-equivalently encoded as:
+
+# RELIABLE_DATAGRAM HTTP/3 Frame Definition {#reliable-datagram-frame}
```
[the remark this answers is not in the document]
afrind: Are datagrams coming to H1?

draft-ietf-masque-h3-datagram.md
```diff
@@ -181,74 +181,82 @@ all cases, the maximum permitted value of the H3_DATAGRAM SETTINGS parameter is
...
+
+The RELIABLE_DATAGRAM frame (type=TBD) is used to send datagrams over QUIC
+streams when QUIC datagrams are unavailable or undesirable. Datagrams
+transmitted over streams using this frame have the same semantics as datgrams
+sent over the QUIC DATAGRAM frame.
```
afrind, line 227: I don't think you mean 'sent over...' a frame?

draft-ietf-masque-h3-datagram.md
```diff
@@ -181,74 +181,82 @@ all cases, the maximum permitted value of the H3_DATAGRAM SETTINGS parameter is
...
 
 ~~~
-  Datagram-Flow-Id = 44; ecn-ect0, 42, 48; ecn-ce, 46; ecn-ect1
+RELIABLE_DATAGRAM Frame {
```
afrind, line 230: What kinds of streams is this frame valid on?  If a flow ID can outlive a stream, how do you send it?

ianswett:
@afrind For QPACK, a table is explicitly maintained, so I think things are Ok.  For h2, it's worse because the HPACK dictionary is LRU-based so if an extension frame is not supported, the dictionary gets out of sync.

afrind:
@ianswett : to clarify, you mean the QPACK table is not modified via encoded field sections.  Still, any extension that makes use of the QPACK dynamic table but doesn't properly decode all the sections will lead to state mismatch between the encoder and decoder -- the decoder still needs to send Section Acknowledgement or Stream Cancellation.  I guess it's not as bad as HPACK. 

I maintain that it's possible to write a safe extension to H2 or H3 that uses HPACK or QPACK dynamic tables, so long as it's guarded under SETTINGS.

draft-ietf-masque-h3-datagram.md
LPardue: I don't know what these frames are in relation to HTTP semantics. They contain an encoded field section but to where does that field section apply? There are rules and documented limitiations for handling trailers https://httpwg.org/http-core/draft-ietf-httpbis-semantics-latest.html#trailers.limitations.

This readslike a weird form of "middlers".

Mixing the use of a compression context between these HEADERS and extensions might be ok, or it might be horribly insecure. Is there precidence for extensions that have done this before?

Maybe extensions can use QPACK but I want to know if H3_DATAGRAM needs to eat that complexity.

DavidSchinazi: Based on this discussion, QPACK isn't as simple as I originally thought, so I replaced it with a simple custom encoding.

afrind: Hm, I think QPACK ended up perhaps too tightly coupled to HTTP semantics - at least it references Field Lines and Field Sections.  The specification itself is otherwise devoid of semantic meaning - it's just an encoding and compression scheme for name/value pairs.  I think it would be kind of a bummer if it couldn't be used, particularly by HTTP extensions, to convey compressed, structured metadata - like "middlers", etc.

But I can see @DavidSchinazi giving us a nudge to quit discussing QPACK here...

LPardue: Oh I'm in complete agreement that HPACK/QPACK are very nice generic key-value compression schemes. It's slightly sad they are entwined with messages but that's what we've been dealt.

draft-ietf-masque-h3-datagram.md
```diff
@@ -181,74 +181,91 @@ all cases, the maximum permitted value of the H3_DATAGRAM SETTINGS parameter is
...
+  ext-member       = ext-member-key "=" ext-member-value
+  ext-member-key   = token
+  ext-member-value = token
 ~~~
 
-In this example, 42 is the unnamed flow, 44 represents the name
-"ecn-ect0", 46 represents "ecn-ect1", and 48 represents "ecn-ce". Note that,
-since the list ordering does not carry semantics, this example can be
-equivalently encoded as:
+Note that these registrations are unilateral and unidirectional: the sender of
```
vasilvv, line 224: This feels weird, because this requires a party to make unilateral declarations about peer's flow IDs.  Why not just make the requester associate a flow from its side to a stream?

draft-ietf-masque-h3-datagram.md
```diff
@@ -181,74 +181,91 @@ all cases, the maximum permitted value of the H3_DATAGRAM SETTINGS parameter is
...
-{{STRUCT-FIELD}}). Each name MUST NOT appear more than once in the list. The
-value of the first parameter of each named element (whose corresponding key
-conveys the element name) MUST be of type Boolean and equal to true. The value
-of the first parameter of the unnamed element MUST NOT be of type Boolean. The
-ordering of the list does not carry any semantics. For example, an HTTP method
-that wishes to use four datagram flows for the lifetime of its
-request stream could look like this:
+: The flow ID to associate with the QUIC stream that carried this frame.
+
+Extension String:
```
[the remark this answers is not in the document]
DavidSchinazi: I'd rather define this once instead of having each method dictate what this contains. Binary would work too, text just felt simpler.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: The issue there is that you would then need some sort of acknowledgement to know when your peer has parsed that frame. The unilateral model removes the need for acknowledgements at the HTTP level (for example HTTP/2 has an ACK bit on the SETTINGS frame for this reason).

draft-ietf-masque-h3-datagram.md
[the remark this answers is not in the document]
LPardue: I think something binary like Settings parameters has been too restrictive for extensibility. Depending exactly what extensibility people migh need, I think I'd probably still lean towards this being an encoded sf-list. This solves for many caveats of data-types and would give a nice pathway to migrate to binary Structured Fields if people really can't stomach strings in wire format.

draft-ietf-masque-h3-datagram.md
afrind: > the sender of the frame unilateraly defines the semantics it will apply to the datagrams it
receives

This feels backwards to me - my inclination is to have the sender define the semantics of datagrams it will send on a flow ID.  This is similar to how unidirectional extension streams work -- the sender-chosen stream type identifies the semantics.

DavidSchinazi: I don't think the analogy with stream types is correct. The semantics of stream types are determined by an IANA registry, whereas flow IDs are ephemeral per-connection numbers. If you reverse the direction then things break in the presence of reordering.

afrind: "I'm listening for X on Flow-ID Y" requires a roundtrip before the sender can even send.  Saying, "I'm sending you X on Flow-ID Y" doesn't break in the face of reordering, you just might lose some datagrams if your stack discards datagrams with unknown Flow IDs.  The receiver can also buffer datagrams on unknown Flow IDs for a while if it's concerned about this case.

Consider my extension for unreliable HTTP upload, where I send a POST request headers on a stream and then the body comes in DATAGRAMs. If I've already seen the SETTINGS from the server and know it supports my extension, why do I need to wait another RTT to acquire a flow-ID?

DavidSchinazi:

PR #38, opened 2021-04-11
Experiment with two layers of indirection

DavidSchinazi:
Similar to #37, this PR is a thought experiment aimed at seeing whether an alternate design makes sense. We're not planning on merging this as-is.

This design was inspired by a proposal by @bemasc. Ben gets the credit for the idea of using two layers of indirection here, and I take the blame for writing it down wrong.

Since this is really hard to review in PR form, I've uploaded an HTML version [here](https://davidschinazi.com/stuff/draft-ietf-masque-h3-datagram.html) so I definitely suggest reading that instead.

Conceptually, I think this proposal is easier to reason about, and it has the advantage of requiring minimal work from intermediaries. Its downside is that the wire format is slightly less efficient because it requires another varint to be sent per datagram, but I think we can live with that.

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -35,10 +35,8 @@ with a mechanism to send unreliable data while leveraging the security and
 congestion-control properties of QUIC. However, QUIC DATAGRAM frames do not
 provide a means to demultiplex application contexts. This document describes how
 to use QUIC DATAGRAM frames when the application protocol running over QUIC is
-HTTP/3. It defines logical flows identified by a non-negative integer that are
-present at the start of the DATAGRAM frame payload. Flows are associated with
-HTTP messages using the Datagram-Flow-Id header field, allowing endpoints to
-match unreliable DATAGRAMS frames to the HTTP messages that they are related to.
+HTTP/3. It associates datagrams with client-initiated bidirectional streams
+and defines an optional additional demultiplexing layer.
 
 Discussion of this work is encouraged to happen on the MASQUE IETF mailing list
 ([masque@ietf.org](mailto:masque@ietf.org)) or on the GitHub repository which
@@ -56,11 +54,8 @@ mechanism to send unreliable data while leveraging the security and
 congestion-control properties of QUIC. However, QUIC DATAGRAM frames do not
 provide a means to demultiplex application contexts. This document describes how
 to use QUIC DATAGRAM frames when the application protocol running over QUIC is
-HTTP/3 {{!H3=I-D.ietf-quic-http}}. It defines logical flows identified by a
-non-negative integer that are present at the start of the DATAGRAM frame
-payload. Flows are associated with HTTP messages using the Datagram-Flow-Id
-header field, allowing endpoints to match unreliable DATAGRAMS frames to the
-HTTP messages that they are related to.
+HTTP/3 {{!H3=I-D.ietf-quic-http}}. It associates datagrams with client-initiated
+bidirectional streams and defines an optional additional demultiplexing layer.
 
 This design mimics the use of Stream Types in HTTP/3, which provide a
 demultiplexing identifier at the start of each unidirectional stream.
@@ -79,49 +74,41 @@ document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
 when, and only when, they appear in all capitals, as shown here.
 
 
-# Datagram Flows
+# Multiplexing
 
-Flows are bidirectional exchanges of datagrams within a single QUIC connection.
-These are conceptually similar to streams in the sense that they allow
-multiplexing of application data. Flows are identified within a connection by a
-numeric value, referred to as the flow ID.  A flow ID is a 62-bit integer (0
-to 2^62-1).
-
-Flows lack any of the ordering or reliability guarantees of streams. Beyond
-this, a sender SHOULD ensure that DATAGRAM frames within a single flow are
-transmitted in order relative to one another. If multiple DATAGRAM frames can be
-packed into a single QUIC packet, the sender SHOULD group them by flow to
-promote fate-sharing within a specific flow and improve the ability to process
-batches of datagram messages efficiently on the receiver.
+In order to allow multiple datagram types and contexts to coexist on a given
+QUIC connection, HTTP datagrams contain two layers of multiplexing. First, the
+QUIC DATAGRAM frame payload starts with a custom-encoded stream identifier that
+associates the datagram with a given QUIC stream. Second, datagrams carry a
+flow ID that allows multiplexing multiple datagram flows related to a given
+HTTP request. Conceptually, the first layer of multiplexing is per-hop, while
+the second is end-to-end (see {{datagram-flows}}).
 
 
-# Flow ID Allocation {#flow-id-alloc}
+# Datagram Flows {#datagram-flows}
+
+Flows are unirectional exchanges of datagrams associated with a given HTTP
+request. Flows are identified within a request context by a numeric value,
+referred to as the flow ID. A flow ID is a 62-bit integer (0 to 2^62-1). Note
+that the flow ID value of zero is reserved for registration purposes (see
+{{register}}).
+
+While stream IDs are a per-hop concept, flow IDs are an end-to-end concept. In
+other words, if a datagram travels through multiple intermediaries on its way
+from client to server, the stream ID will most likely change from hop to hop,
+but the flow ID will remain the same. Flow IDs are opaque to intermediaries.
+
+
+## Flow ID Allocation {#flow-id-alloc}
 
 Implementations of HTTP/3 that support the DATAGRAM extension MUST provide a
 flow ID allocation service. That service will allow applications
 co-located with HTTP/3 to request a unique flow ID that they can
 subsequently use for their own purposes. The HTTP/3 implementation will then
 parse the flow ID of incoming DATAGRAM frames and use it to deliver the
-frame to the appropriate application context.
-
-Even-numbered flow IDs are client-initiated, while odd-numbered flow
-IDs are server-initiated. This means that an HTTP/3 client
-implementation of the flow ID allocation service MUST only provide
-even-numbered IDs, while a server implementation MUST only provide
-odd-numbered IDs. Note that, once allocated, any flow ID can be
-used by both client and server - only allocation carries separate namespaces to
-avoid requiring synchronization.
-
-The flow ID allocation service SHOULD also provide a mechanism for applications
-to indicate they have completed their usage of a flow ID and will no
-longer be using that flow ID, this process is called "retiring" a flow
-ID. Applications MUST NOT retire a flow ID until after they
-have received confirmation that the peer has also stopped using that flow
-ID. The flow ID allocation service MAY reuse previously
-retired flow ID once they have ascertained that there are no packets
-with DATAGRAM frames using that flow ID still in flight. Reusing flow
-IDs can improve performance by transmitting the flow ID using
-a shorter variable-length integer encoding.
+frame to the appropriate application context. Note that the flow ID namespace is
+tied to a given HTTP request: it is possible for the same numeral flow ID to be
+used simultaneously in distinct requests.
 
 
 # HTTP/3 DATAGRAM Frame Format {#format}
@@ -132,11 +119,20 @@ of {{QUIC}}):
 
 ~~~
 HTTP/3 DATAGRAM Frame {
+  Quarter Stream ID (i),
   Flow ID (i),
   HTTP/3 Datagram Payload (..),
 }
 ~~~
 {: #h3-datagram-format title="HTTP/3 DATAGRAM Frame Format"}
+
+Quarter Stream ID:
+
+: A variable-length integer that contains the value of the client-initiated
+bidirectional stream that this datagram is associated with, divided by four.
+(The division by four stems from the fact that HTTP requests are sent on
+client-initiated bidirectional streams, and those have stream IDs equal to zero
+modulo four.)
 
 Flow ID:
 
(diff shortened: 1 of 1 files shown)


LPardue:
I'm still digesting this proposal, so take the following comments with a pinch of salt.

I think the PR description glosses over some design changes and buries the lede. The major changes in this proposal seems to be that Flow IDs are no longer connection global, and Flow ID usage negotitation is no longer based on headers. If I were to reword the proposed concepts based on my understanding, we're talking about flow groups bound to each request. Flows can be added to the group during its lifetime but never retired. A flow allocation service needs only ensure uniqueness of flow IDs in a flow group. I'm failing to see why it is better to have an identifying tuple (stream-id:flow-id) instead of a global identifier (flow-id) but happy to be pointed out to why.

The examples are helpful so thanks for putting those in (I might have opinions on actually keeping them but this spec is currently weaker for not including any, I think we tracked this on a ticket already.)

If I were to summarise my understanding, this proposal replaces

```
HEADERS (on stream 44) {
  method: CONNECT-UDP
  {other headers}
  datagram-flow-id: 0
}
```

with 

```
HEADERS (on stream 44) {
  method: CONNECT-UDP
  {other headers}
}

RELIABLE_DATAGRAM (on stream 44) { 
  Flow-ID: 0
  Registered-Flow-Id: 1
  Extension-String: ""
}
```

The claim is that this makes it easier to reason about things, but frankly at this stage I can see it making some things harder. No longer can a MASQUE server reason about creating a tunnel based on HEADERS alone, it needs to combine that with some additional information that is delivered on the request stream but is not part of an HTTP request semantic. Such is the nature of H2/H3 frames, but I find it odd that the flow ID is described as an end-to-end signal in the same breath as being described as bound to a request context.

The surprising change I find in this design is it makes the flow registration unreliable, with a very weak suggestion to use something reliable. Can you embelish on why you picked this direction?  I'm not convinced if we want to do this flow groups thing but if we did, I find the overloading of RELIABLE_DATAGRAM unhelpful. It seems that you could combine some of this proposal with the REGISTER_DATAGRAM_FLOW_ID from #37. That would provide a clearer separation between the datagram control plane and data plane. One advantage of using a frame delivered on the request stream is that it avoids Flow-ID 0 being special case, which is ugly because it prevents requets from being made on Stream-ID 0. Another advantage is that it allows strict ordering of frames, in case we wanted to add rules about in-order flow ID usage.

Finally, I'm slightly confused by how the CONNECT-IP example would work in reality if the server did not support the `"ip=192.0.2.42,port=443"` extension. If it drops the DATAGRAM payload silently, that seems like a sad move. I think someone (maybe Alan?) suggested somewhere that a frame-based registration design could benefit from some control-plane signal. I'm beginning to lean toward some further types of controls for flows, like an H3-layer STOP_SENDING/RESET_FLOW.

To to conclude my understanding/opinion of this proposal

1. flow IDs are request-scoped, not global. I'm not sure if this is a good, bad or neutral change.
1. flow IDs reservation is  done using a frame and can be done during the lifetime of a request
  * This makes things gnarlier for my current implementation. It would require API changes to expose such events to an application.
1. flow ID reservation could include extension strings I don't understand. If so, I have to just drop all datagrams for that flow and can't signal it back to my peer.

bemasc:
Hi Lucas,

>No longer can a MASQUE server reason about creating a tunnel based on HEADERS alone, it needs to combine that with some additional information that is delivered on the request stream but is not part of an HTTP request semantic.

I'm not sure what you mean by "creating a tunnel", but note that the actual call to `bind()`, to allocate a UDP socket, can indeed be performed based on the headers alone.  (Actually, you don't even need the headers.)

It's true that, like #37, this proposal allows flows to be created at any time during CONNECT-UDP's lifetime.  This feature is not strictly necessary, but can be useful in various scenarios.  For example, it is hard to imagine how QUIC connection ID compression can work without this.

> Such is the nature of H2/H3 frames, but I find it odd that the flow ID is described as an end-to-end signal in the same breath as being described as bound to a request context.

HTTP requests are end-to-end: they are communicated through the intermediaries but are not _for_ the intermediaries.  They're not even version-specific.  Flows in MASQUE have these properties too.

> The surprising change I find in this design is it makes the flow registration unreliable, with a very weak suggestion to use something reliable. Can you embelish on why you picked this direction?

I interpreted this as "you definitely want to use reliable registration, but technically nobody is going to enforce this requirement on you".  I think we can adjust this easily enough.

> I'm not convinced if we want to do this flow groups thing but if we did, I find the overloading of RELIABLE_DATAGRAM unhelpful. It seems that you could combine some of this proposal with the REGISTER_DATAGRAM_FLOW_ID from #37. That would provide a clearer separation between the datagram control plane and data plane.

The key claim of this design is that flows are only meaningful end-to-end, so exposing them to intermediaries is unnecessary.  From the HTTP transport's perspective, a flow-registration is just more data to be routed along with the request/response; it's only meaningful to the client and server.

> One advantage of using a frame delivered on the request stream is that it avoids Flow-ID 0 being special case, which is ugly because it prevents requets from being made on Stream-ID 0.

FWIW, I think "zero is special" is a fairly elegant common pattern.

> Another advantage is that it allows strict ordering of frames, in case we wanted to add rules about in-order flow ID usage.

It's not spelled out here, but RELIABLE_DATAGRAMs would be strictly ordered, so this is still possible.

> Finally, I'm slightly confused by how the CONNECT-IP example would work in reality if the server did not support the `"ip=192.0.2.42,port=443"` extension.

I think @DavidSchinazi is implying that the listed extensions all happen to be Mandatory To Implement for the corresponding methods.  I agree, it would be good to discuss how non-mandatory extensions are negotiated.

> I'm beginning to lean toward some further types of controls for flows, like an H3-layer STOP_SENDING/RESET_FLOW.

This is a key advantage of the two-layer design here.  These types of commands are only meaningful end-to-end, but without the second layer, adding new commands requires upgrading every intermediary and negotiating support in SETTINGS at each hop.

LPardue:
Hey Ben, 

Thanks for the reply, it paints a much fuller picture. I'm still wondering how this form of design might work across HTTP/1, 2 and 3 but that is a broader question for the WG so I'll leave aside. I'm not sure how best to reply here but I'll give quoting a shot. 

> I'm not sure what you mean by "creating a tunnel", but note that the actual call to bind(), to allocate a UDP socket, can indeed be performed based on the headers alone. (Actually, you don't even need the headers.)

As my first comment, I probably had the weakest view of the proposal. Based on the current design, a MASQUE server could reject a request if, for example, it detected a bad flow ID. I realize now that some of those failure conditions just go away.

> It's true that, like #37, this proposal allows flows to be created at any time during CONNECT-UDP's lifetime. This feature is not strictly necessary, but can be useful in various scenarios. For example, it is hard to imagine how QUIC connection ID compression can work without this.

Yep this seems like a useful extension point and something that is difficult to achieve today.

> >     The surprising change I find in this design is it makes the flow registration unreliable, with a very weak suggestion to use something reliable. Can you embelish on why you picked this direction?

> I interpreted this as "you definitely want to use reliable registration, but technically nobody is going to enforce this requirement on you". I think we can adjust this easily enough.

Why not enforce it though? What scenario benefits from an unreliable flow creation. DATAGRAMs can be lost in the network or in the application buffers. The opportunity for problems get worse when adding intermediaries.

> The key claim of this design is that flows are only meaningful end-to-end, so exposing them to intermediaries is unnecessary. From the HTTP transport's perspective, a flow-registration is just more data to be routed along with the request/response; it's only meaningful to the client and server.

That explanation helps, thanks. I think this speaks to my earlier point though. It seems this is a design that adds stuff to HTTP message but is not HTTP metadata or HTTP content. So what is it? H2 and H3 frames are hop-by-hop but their contents can be forwarded without being processed by anyone but the intended recipient. An intermediary that knows how to forward a request can forward frame contents. It seems this is the desired goal of using RELIABLE_DATAGRAM for control and data planes. But my point is that you could equally define REGISTER_DATAGRAM_FLOW_ID and other control frames that behave similarly, without having to overload them as a DATAGRAM on a special flow.

> FWIW, I think "zero is special" is a fairly elegant common pattern.

We avoided it for QUIC and HTTP/3. If every deployment of HTTP/3 DATAGRAM would have to burn the first request stream, that seems ugly to me.

>>    Finally, I'm slightly confused by how the CONNECT-IP example would work in reality if the server did not support the "ip=192.0.2.42,port=443" extension.

> I think @DavidSchinazi is implying that the listed extensions all happen to be Mandatory To Implement for the corresponding methods. I agree, it would be good to discuss how non-mandatory extensions are negotiated.

>>    I'm beginning to lean toward some further types of controls for flows, like an H3-layer STOP_SENDING/RESET_FLOW.

> This is a key advantage of the two-layer design here. These types of commands are only meaningful end-to-end, but without the second layer, adding new commands requires upgrading every intermediary and negotiating support in SETTINGS at each hop.

I see two separate things. First, it is useful to be able to handle DATAGRAM frames with different formats, without needing all points in the change to understand how to process that format - I agree with this. Second, we can use DATAGRAM frames to carry control messages for managing datagram flows. This one I have more of a problem with, how extensible does flow management need to be? For this proposal, changes are needed to add RELIABLE_DATAGRAM support through the chain in the first instance. Adding support for other frames at the same time doesn't much more difficult.

bemasc:
> > > ```
> > > The surprising change I find in this design is it makes the flow registration unreliable, with a very weak suggestion to use something reliable. Can you embelish on why you picked this direction?
> > > ```
> 
> > I interpreted this as "you definitely want to use reliable registration, but technically nobody is going to enforce this requirement on you". I think we can adjust this easily enough.
> 
> Why not enforce it though?

I'm fine with enforcing it, so long as we're clear that enforcement only happens at the endpoints.  Intermediaries shouldn't be looking past the Quarter Stream ID.

> > The key claim of this design is that flows are only meaningful end-to-end, so exposing them to intermediaries is unnecessary. From the HTTP transport's perspective, a flow-registration is just more data to be routed along with the request/response; it's only meaningful to the client and server.
> 
> That explanation helps, thanks. I think this speaks to my earlier point though. It seems this is a design that adds stuff to HTTP message but is not HTTP metadata or HTTP content. So what is it?

In an idealized version of HTTP, I see a stack of layers: hop, cache, exchange (request/response, end-to-end), method, resource (payload/representation).  The Quarter Stream ID is part of the hop layer.  Everything past that point is owned by the exchange.  The Flow ID, commands, and extension types are at the method layer.  The content of the extensions is at the resource layer.

Concretely, this means that intermediaries can forward datagrams for unrecognized methods, even if we cannot predict what requirements future methods may have for datagrams.

> H2 and H3 frames are hop-by-hop but their contents can be forwarded without being processed by anyone but the intended recipient.

Yes, and this PR essentially defines a frame type that means "forward this without processing unless you're the intended recipient".

> An intermediary that knows how to forward a request can forward frame contents. It seems this is the desired goal of using RELIABLE_DATAGRAM for control and data planes. But my point is that you could equally define REGISTER_DATAGRAM_FLOW_ID and other control frames that behave similarly, without having to overload them as a DATAGRAM on a special flow.
> 
> > FWIW, I think "zero is special" is a fairly elegant common pattern.
> 
> We avoided it for QUIC and HTTP/3. If every deployment of HTTP/3 DATAGRAM would have to burn the first request stream, that seems ugly to me.

The obvious alternative is to put a DatagramType varint after the Quarter Stream ID, like an H3 frame on a stream.  This costs us one byte of MTU.  I don't think it buys us anything concrete, but I'm willing to spend a byte if it makes the design feel more consistent.

> > > I'm beginning to lean toward some further types of controls for flows, like an H3-layer STOP_SENDING/RESET_FLOW.
> 
> > This is a key advantage of the two-layer design here. These types of commands are only meaningful end-to-end, but without the second layer, adding new commands requires upgrading every intermediary and negotiating support in SETTINGS at each hop.
> 
> I see two separate things. First, it is useful to be able to handle DATAGRAM frames with different formats, without needing all points in the change to understand how to process that format - I agree with this. Second, we can use DATAGRAM frames to carry control messages for managing datagram flows. This one I have more of a problem with, how extensible does flow management need to be? For this proposal, changes are needed to add RELIABLE_DATAGRAM support through the chain in the first instance. Adding support for other frames at the same time doesn't much more difficult.

Exposing the concept of flows to intermediaries makes them everything much more complicated.  For example, intermediaries would need their own flow count limits to avoid memory exhaustion.  Will clients and servers be able to figure out what happened when that limit is reached?  In the case of flow cleanup commands, the commands would presumably need to be ack'd end-to-end before flows could be reused, and both the commands and the acks would need to be interpreted and translated by intermediaries.  As an intermediary, what do I do with a datagram on a mapping for which a "stop" command has been forwarded but the "ack" has not yet been observed?  This complexity is not buying us anything.

Extensibility, with additional commands negotiated via SETTINGS, is even more of a mess.  If I'm a CDN reverse proxy for myriad backends with different capabilities, what commands do I claim to support?

I think we're much better off if we can draw a clear, simple line between hop-by-hop and end-to-end data.

LPardue:
Just to respond to one specific point:

>>    I see two separate things. First, it is useful to be able to handle DATAGRAM frames with different formats, without needing all points in the change to understand how to process that format - I agree with this. Second, we can use DATAGRAM frames to carry control messages for managing datagram flows. This one I have more of a problem with, how extensible does flow management need to be? For this proposal, changes are needed to add RELIABLE_DATAGRAM support through the chain in the first instance. Adding support for other frames at the same time doesn't much more difficult.

> Exposing the concept of flows to intermediaries makes them everything much more complicated. For example, intermediaries would need their own flow count limits to avoid memory exhaustion. Will clients and servers be able to figure out what happened when that limit is reached? In the case of flow cleanup commands, the commands would presumably need to be ack'd end-to-end before flows could be reused, and both the commands and the acks would need to be interpreted and translated by intermediaries. As an intermediary, what do I do with a datagram on a mapping for which a "stop" command has been forwarded but the "ack" has not yet been observed? This complexity is not buying us anything.

I think we might be talking across ourselves but mean the same thing. All I'm suggesting is that a REGISTER_DATAGRAM_FLOW_ID (or whatever we call it) could contain the same information as the proposed RELIABLE_DATAGRAM on flow 0. Intermediaries wouldn't process the contents of the frame, they'd just forward it on the stream.

to repurpose an example


```
~~~
Client                                             Server

STREAM(44): DATA{HEADERS}      -------->
  :method = CONNECT-UDP
  :scheme = https
  :path = /
  :authority = target.example.org:443

STREAM(44): REGISTER_DATAGRAM_FLOW_ID  -------->
  Registered Flow ID = 1
  Extension String = ""

DATAGRAM                       -------->
  Quarter Stream ID = 11
  Flow ID = 1
  Payload = Encapsulated UDP Payload

           <--------  STREAM(44): DATA{HEADERS}
                        :status = 200

           <--------  STREAM(44): REGISTER_DATAGRAM_FLOW_ID
                        Registered Flow ID = 1
                        Extension String = ""

/* Wait for target server to respond to UDP packet. */

           <--------  DATAGRAM
                        Quarter Stream ID = 11
                        Flow ID = 1
                        Payload = Encapsulated UDP Payload
~~~
```

bemasc:
Got it, thanks.  I agree, that works fine.  However, I do feel that it's somewhat less elegant and much less extensible.  Ideally, I would like all data that intermediaries don't need to process to be opaque to them, so there's no ambiguity about what is hop-by-hop and what is end-to-end.  Or to put it another way: at the transport level, "REGISTER_DATAGRAM_FLOW_ID" and "RELIABLE_DATAGRAM" have the same semantics, so why define it twice?

Another way to see this is to consider HTTP/1.1 usage.  In that case, we literally need to encode the flow registration commands into the request and response bodies.  That makes it clear that these commands are part of the end-to-end resource.

LPardue:
IIUC this proposal moves the flow ID registration from an always reliable transfer, to a maybe reliable one. I think that's important to call out and get group consensus. If am understanding incorrectly, then please do correct me :)

Generally we're carry traffic intended for the target server in datagrams, and we expect that tunneled protocol to run it's own loss detection algorithm if it cares. But flow 0 here is special case and is seems like it might need it's own loss detection too.

At risk of going off on a tangent, as an implementer I need to decide how to handle RELIABLE_DATAGRAM. Is it reliable just for the hop that it is being carried over, or should it be reliable all the way? This PR implies its ok to leave the reliability choice up to implementations, and I don't trust them not to have hard-to-debug problems. I.e. say we had

masque client -> intermediary A -> intermediary B -> masque server

If client sent RELIABLE_DATAGRAM to A, is A allowed to shift the contents into a DATAGRAM for transmission to B?
Or what if I write an implementation that operates a shared in-memory queue of DATAGRAMS for all flows, that drops the oldest to maintain a fixed size. Am I allowed to put RELIABLE_DATAGRAMs on that queue?

LPardue:
A different point, before I forget. I like this proposal's change of flow IDs to be unidirectional. The Flow ID becomes more about the format of the datagram message, rather than it's purpose, which is powerful. The CONNECT-IP example is easy to overlook, the client is able to send DATAGRAMS either uncompressed or compressed, while the server can only do uncompressed. 

The impact to an implementation that handles the DATAGRAM is that is needs to keep 2 sets of {stream:flow} tuples for each stream - one in the send direction, one in the receive direction. But I think that is manageable.

bemasc:
This PR doesn't spell out RELIABLE_DATAGRAM, but you're right that the details there are important.  In H2, RELIABLE_DATAGRAM needs to carry a flag to ensure that messages that need to be reliable (like control messages) stay reliable end-to-end, and are not converted into H3 unreliable datagrams at H2->H3 intermediaries.

DavidSchinazi:

PR #39, opened 2021-04-12
Yet another experiment with two layers of indirection

DavidSchinazi:
This is a variation on #38 but with the "flow ID = 0 means register" replaced with the REGISTER_DATAGRAM_FLOW_ID frame from #37.

Since this is really hard to review in PR form, I've uploaded an HTML version [here](https://davidschinazi.com/stuff/draft-ietf-masque-h3-datagram-exp2.html#name-register_datagram_flow_id-ht) so I definitely suggest reading that instead.

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -35,10 +35,8 @@ with a mechanism to send unreliable data while leveraging the security and
 congestion-control properties of QUIC. However, QUIC DATAGRAM frames do not
 provide a means to demultiplex application contexts. This document describes how
 to use QUIC DATAGRAM frames when the application protocol running over QUIC is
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-h3-datagram.md
```diff
@@ -182,102 +220,49 @@ all cases, the maximum permitted value of the H3_DATAGRAM SETTINGS parameter is
 1.
 
 
-# Datagram-Flow-Id Header Field Definition {#header}
+# RELIABLE_DATAGRAM HTTP/3 Frame Definition {#reliable-datagram-frame}
```
tfpauly, line 223: I see why this frame is necessary for #38, since that was used to register flows, but I'd prefer to see this removed from this version of the proposal for simplicity and clarity. If people need it, it can be an extension later.

DavidSchinazi:

Issue #45, opened 2021-04-16
Is the context ID optional?


DavidSchinazi:
This issue assumes that decide to go with the two-layer design described in #41. Given that design, some applications might not need the multiplexing provided by context IDs. We have multiple options here:

* Make context ID mandatory
  * Applications that don't need it waste a byte per datagram
* Negotiate the presence of context IDs using an HTTP header
  * Context IDs would still be mandatory to implement on servers because the client might send that header
* Have the method determine whether context IDs are present or not
  * This would prevent extensibility on those methods
* Create a REGISTER_NO_CONTEXT message
  * This assumes we use the Message design from #44
  * We add a register message that means that this stream does not use context IDs
  * This allows avoiding the byte overhead without sacrificing extensibility

tfpauly:
Making the context ID mandatory is simple, although slightly wasteful.

For the variations where it is optional, I prefer the one that uses a header. It could work with a variant of #44 where we do Header + Message, and the Header determines whether or not you have contexts, and whether or not they can be dynamic.

achernya:
What are the benefits of message over header to make it optional? To me it seems like header is sufficient, because the context IDs are end-to-end, and should not affect intermediaries, making the header a natural place to put it. I think having the context ID be the default behavior and the header is to opt-out is the right decision if we were to allow removing it.

afrind:
>  I think having the context ID be the default behavior and the header is to opt-out is the right decision if we were to allow removing it.

This is a bit of a bike-shed, but why would it default to having context-IDs instead of not having them?

ianswett:
I think the second option "Negotiate the presence of context IDs using an HTTP header" is the best personally.

In terms of: "Context IDs would still be mandatory to implement on servers because the client might send that header", I assume server in this context is the final application server, not a proxy?  Even so, I'd argue it's applications which decide if they need this functionality, so if a server only implemented one application and it didn't use Context-IDs, then I don't see why it would need to implement it.

achernya:
Ian, in your comment, are you talking about an intermediary when you say "proxy"? If so, then an intermediary need not implement anything based on the payload of than H3 DATAGRAM, it only has to pass the data along. That may be worth a clarification.

LPardue:
Having context ID optional has a nice future-proof property - other uses of datagrams in HTTP/3 can decide if they want the second layer or not. 

[...]

PR #52, opened 2021-05-04
Two layer design with CAPSULE

### LINKED ISSUE #41: 2-layer design
During the 2021-04 MASQUE Interim, we discussed a design proposal from @bemasc to replace the one-layer flow ID design with a two-layer design. This design replaces the flow ID with two numbers: a stream ID and a context ID. The contents of the QUIC DATAGRAM frame would now look like:
```
HTTP/3 DATAGRAM Frame {
  Stream ID (i),
  Context ID (i),
  HTTP/3 Datagram Payload (..),
}
```

While the flow ID in the draft was per-hop, the proposal has better separation:
* the stream ID is per-hop, it maps to an HTTP request
* the context ID is end-to-end, it maps to context information inside that request

Intermediaries now only look at stream IDs and can be blissfully ignorant of context IDs.

In the room at the 2021-04 MASQUE Interim, multiple participants spoke in support of this design and no one raised objections. This issue exists to ensure that we have consensus on this change - please speak up if you disagree.

### LINKED ISSUE #42: Directionality of flow/context IDs
In draft-ietf-masque-h3-datagram-00, flow IDs are bidirectional. During the 2021-04 MASQUE Interim, we discussed the possibility of making them unidirectional. Here are the differences:

* Bidirectional
  * single shared namespace for both endpoints (even for client, odd for server)
  * it's easier to refer to peer's IDs, which makes negotiation easier

* Unidirectional
  * one namespace per direction
  * every bidirectional use needs to associate both directions somehow

I'm personally leaning towards keeping them bidirectional. Unlike QUIC streams, these do not have a state machine or flow control credits, so there is no cost to only using a single direction of a bidirectional flow for unidirectional use-cases.

### LINKED ISSUE #43: Ability to accept or reject registration of context IDs
During the 2021-04 MASQUE Interim, we discussed having the ability to accept or reject registration of context IDs. An important motivation for this is the fact that context ID registrations take up memory (for example it could be a compression context which contains the data that is elided) and therefore we need to prevent endpoints from arbitrarily making their peer allocate memory. While we could flow control context ID registration, a much simpler solution would be to have a way for endpoints to close/shutdown a registration.

### LINKED ISSUE #44: Pick mechanism to register context IDs (header vs message)
During the 2021-04 MASQUE Interim, we discussed multiple options for registering context IDs. Without going into the specific of the encoding, there are two broad classes of solutions:

* "Header" design
  * Once-at-setup registration
  * This happens during the HTTP request/response
  * Simplest solution is to use an HTTP header such as `Datagram-Flow-Id` (encoding TBD)
  * Example use-case: CONNECT-UDP without extensions
* "Message" design
  * Mid-stream registration
  * This can happen at any point during the lifetime of the request stream
  * This would involve some sort of "register" message (encoding TBD, see separate issue)
  * Example use-case: CONNECT-IP on-the-fly compression (this can't be done at tunnel setup time because the bits top be compressed aren't always known at that time)
 
It's possible to implement once-at-setup registration using the Message design, but it isn't possible to implement mid-stream registration. Therefore I think we should go with the Message design. It would also be possible to implement both but I don't think that provides much value.

DavidSchinazi:
This PR aims to take the design discussions we had at the last
interim, and propose a potential solution. This PR should in theory
contain enough information to allow folks to implement this new
design. That will allow us to perform interop testing and confirm
whether the design is sound.

[An HTML version of the draft with these changes is available here.](https://ietf-wg-masque.github.io/draft-ietf-masque-h3-datagram/two_layer_capsule/draft-ietf-masque-h3-datagram.html)

Closes #41, closes #42, closes #43, closes #44.

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -186,7 +186,8 @@ the presence of HTTP intermediaries.
 CAPSULE is an HTTP/3 Frame (as opposed to a QUIC frame) which SHALL only be sent
 in client-initiated bidirectional streams. Intermediaries MUST forward all
 received CAPSULE frames in their unmodified entirety on the same stream where it
-would forward DATA frames.
+would forward DATA frames. Intermediaries MUST NOT send any CAPSULE frames other
+than the ones it is forwarding.
 
 This specification of CAPSULE currently uses HTTP/3 frame type 0xffcab5. If this
 document is approved, a lower number will be requested from IANA.
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -73,24 +73,29 @@ when, and only when, they appear in all capitals, as shown here.
 
 # Multiplexing
 
-In order to allow multiple datagram types and contexts to coexist on a given
-QUIC connection, HTTP datagrams contain two layers of multiplexing. First, the
-QUIC DATAGRAM frame payload starts with a custom-encoded stream identifier that
-associates the datagram with a given QUIC stream. Second, datagrams carry a
-context ID that allows multiplexing multiple datagram contexts related to a
-given HTTP request. Conceptually, the first layer of multiplexing is per-hop,
-while the second is end-to-end.
+In order to allow multiple exchanges of datagrams to coexist on a given QUIC
+connection, HTTP datagrams contain two layers of multiplexing. First, the QUIC
+DATAGRAM frame payload starts with an encoded stream identifier that associates
+the datagram with a given QUIC stream. Second, datagrams carry a context
+identifier (see {{datagram-contexts}}) that allows multiplexing multiple
+datagram contexts related to a given HTTP request. Conceptually, the first
+layer of multiplexing is per-hop, while the second is end-to-end.
 
 
 ## Datagram Contexts {#datagram-contexts}
 
-Contexts refer to bidirectional exchanges of datagrams associated with a single
-HTTP request. Contexts are identified within the scope of a given request by a
-numeric value, referred to as the context ID. A context ID is a 62-bit integer
-(0 to 2^62-1).
+Within the scope of a given HTTP request, contexts provide an additional
+demultiplexing layer. Contexts determine the encoding of datagrams, and can be
+used to implicitly convey metadata. For example, contexts can be used for
+compression to elide some parts of the datagram: the context identifier then
+maps to a compression context that the receiver can use to reconstruct the
+elided data.
+
+Contexts are identified within the scope of a given request by a numeric value,
+referred to as the context ID. A context ID is a 62-bit integer (0 to 2^62-1).
 
 While stream IDs are a per-hop concept, context IDs are an end-to-end concept.
-In other words, if a datagram travels through multiple intermediaries on its
+In other words, if a datagram travels through one or more intermediaries on its
 way from client to server, the stream ID will most likely change from hop to
 hop, but the context ID will remain the same. Context IDs are opaque to
 intermediaries.
@@ -166,6 +171,12 @@ the Quarter Stream ID field but too short to allow parsing the Context ID
 field, the endpoint MUST abruptly terminate the corresponding stream with a
 stream error of type H3_GENERAL_PROTOCOL_ERROR.
 
+If a DATAGRAM frame is received and its Quarter Stream ID maps to a stream that
+has already been closed, the receiver MUST silently drop that frame. If a
+DATAGRAM frame is received and its Quarter Stream ID maps to a stream that has
+not yet been created, the receiver SHALL either drop that frame silently or
+buffer it temporarily while awaiting the creation of the corresponding stream.
+
 
 # CAPSULE HTTP/3 Frame Definition {#capsule-frame}
 
@@ -174,7 +185,7 @@ the presence of HTTP intermediaries.
 
 CAPSULE is an HTTP/3 Frame (as opposed to a QUIC frame) which SHALL only be sent
 in client-initiated bidirectional streams. Intermediaries MUST forward all
-received CAPSULE frame in their unmodified entirety on the same stream where it
+received CAPSULE frames in their unmodified entirety on the same stream where it
 would forward DATA frames.
 
 This specification of CAPSULE currently uses HTTP/3 frame type 0xffcab5. If this
@@ -209,8 +220,8 @@ know the Capsule Type or cannot parse the Capsule Data.
 ## The REGISTER_DATAGRAM_CONTEXT Capsule {#register-capsule}
 
 The REGISTER_DATAGRAM_CONTEXT capsule (type=0x00) allows an endpoint to inform
-its peer of the encoding and semantics of upcoming datagrams associated with
-a given context ID. Its Capsule Data field consists of:
+its peer of the encoding and semantics of datagrams associated with a given
+context ID. Its Capsule Data field consists of:
 
 ~~~
 REGISTER_DATAGRAM_CONTEXT Capsule {
@@ -226,7 +237,8 @@ Context ID:
 
 Extension String:
 
-: A string of comma-separated key-value pairs to enable extensibility.
+: A string of comma-separated key-value pairs to enable extensibility. Keys are
+registered with IANA, see {{iana-keys}}.
 
 The ABNF for the Extension String field is as follows (using syntax from
 {{Section 3.2.6 of RFC7230}}):
@@ -247,8 +259,16 @@ Endpoints MUST NOT send DATAGRAM frames using a Context ID until they have
 either sent or received a REGISTER_DATAGRAM_CONTEXT Capsule with the same
 Context ID. However, due to reordering, an endpoint that receives a DATAGRAM
 frame with an unknown Context ID MUST NOT treat it as an error, it SHALL
-instead drop the DATAGRAM frame silently, or buffer it for a short while in
-hopes of receiving the corresponding REGISTER_DATAGRAM_CONTEXT Capsule.
+instead drop the DATAGRAM frame silently, or buffer it temporarily while
+awaiting the corresponding REGISTER_DATAGRAM_CONTEXT Capsule.
+
+Endpoints MUST NOT register the same Context ID twice on the same stream. This
+also applies to Context IDs that have been closed using a
+CLOSE_DATAGRAM_CONTEXT capsule. Clients MUST NOT register server-initiated
+Context IDs and servers MUST NOT register client-initiated Context IDs. If an
+endpoint receives a REGISTER_DATAGRAM_CONTEXT capsule that violates one or more
+of these requirements, the endpoint MUST abruptly terminate the corresponding
+stream with a stream error of type H3_GENERAL_PROTOCOL_ERROR.
 
 
 ## The CLOSE_DATAGRAM_CONTEXT Capsule {#close-capsule}
@@ -281,6 +301,12 @@ it MUST NOT send any DATAGRAM frames with that Context ID. However, due to
 reordering, an endpoint that receives a DATAGRAM frame with a closed Context ID
 MUST NOT treat it as an error, it SHALL instead drop the DATAGRAM frame
 silently.
+
+Endpoints MUST NOT close a Context ID that was not previously registered.
+Endpoints MUST NOT close a Context ID that has already been closed. If an
+endpoint receives a CLOSE_DATAGRAM_CONTEXT capsule that violates one or more of
+these requirements, the endpoint MUST abruptly terminate the corresponding
+stream with a stream error of type H3_GENERAL_PROTOCOL_ERROR.
 
 
 ## The DATAGRAM Capsule {#datagram-capsule}
@@ -431,8 +457,9 @@ This registry initially contains the following entries:
 
 ## Context Extension Keys {#iana-keys}
 
-This document will request IANA to create an "HTTP Datagram Context
-Extension Keys" registry. Registrations in this registry MUST
+REGISTER_DATAGRAM_CONTEXT capsules carry key-value pairs, see
+{{register-capsule}}. This document will request IANA to create an "HTTP
+Datagram Context Extension Keys" registry. Registrations in this registry MUST
 include the following fields:
 
 Key:
@@ -532,7 +559,7 @@ STREAM(44): CAPSULE            -------->
 DATAGRAM                       -------->
   Quarter Stream ID = 11
   Context ID = 2
-  Payload = Encapsulated UDP Payload With TImeStamp
+  Payload = Encapsulated UDP Payload With Timestamp
 ~~~
 
 
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -79,13 +79,13 @@ QUIC DATAGRAM frame payload starts with a custom-encoded stream identifier that
 associates the datagram with a given QUIC stream. Second, datagrams carry a
 context ID that allows multiplexing multiple datagram contexts related to a
 given HTTP request. Conceptually, the first layer of multiplexing is per-hop,
-while the second is end-to-end (see {{datagram-contexts}}).
+while the second is end-to-end.
 
 
-# Datagram Contexts {#datagram-contexts}
+## Datagram Contexts {#datagram-contexts}
 
-Contexts are birectional exchanges of datagrams associated with a given HTTP
-request. Contexts are identified within the scope of a given request by a
+Contexts refer to bidirectional exchanges of datagrams associated with a single
+HTTP request. Contexts are identified within the scope of a given request by a
 numeric value, referred to as the context ID. A context ID is a 62-bit integer
 (0 to 2^62-1).
 
@@ -136,8 +136,8 @@ Quarter Stream ID:
 : A variable-length integer that contains the value of the client-initiated
 bidirectional stream that this datagram is associated with, divided by four.
 (The division by four stems from the fact that HTTP requests are sent on
-client-initiated bidirectional streams, and those have stream IDs equal to zero
-modulo four.)
+client-initiated bidirectional streams, and those have stream IDs that are
+divisible by four.)
 
 Context ID:
 
@@ -239,7 +239,7 @@ The ABNF for the Extension String field is as follows (using syntax from
 ~~~
 
 Note that these registrations are unilateral and bidirectional: the sender of
-the frame unilateraly defines the semantics it will apply to the datagrams it
+the frame unilaterally defines the semantics it will apply to the datagrams it
 sends and receives using this context ID. Once a context ID is registered, it
 can be used in both directions.
 
@@ -269,9 +269,9 @@ Context ID:
 : The context ID to close.
 
 Note that this close is unilateral and bidirectional: the sender of the frame
-unilateraly informs its peer of the closure. Endpoints can use
+unilaterally informs its peer of the closure. Endpoints can use
 CLOSE_DATAGRAM_CONTEXT capsules to close a context that was initially
-registered by either themself, or by their peer. Endpoints MAY use the
+registered by either themselves, or by their peer. Endpoints MAY use the
 CLOSE_DATAGRAM_CONTEXT capsule to immediately reject a context that was just
 registered using a REGISTER_DATAGRAM_CONTEXT capsule if they find its Extension
 String to be unacceptable.
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -123,7 +123,7 @@ following format (using the notation from the "Notational Conventions" section
 of {{QUIC}}):
 
 ~~~
-HTTP/3 DATAGRAM Frame {
+HTTP/3 Datagram {
   Quarter Stream ID (i),
   Context ID (i),
   HTTP/3 Datagram Payload (..),
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -149,9 +149,22 @@ HTTP/3 Datagram Payload:
 : The payload of the datagram, whose semantics are defined by individual
 applications. Note that this field can be empty.
 
-Endpoints MUST treat receipt of a DATAGRAM frame whose payload is too short to
-parse the Quarter Stream ID or Context ID as an HTTP/3 connection error of type
-H3_GENERAL_PROTOCOL_ERROR.
+Intermediaries parse the Quarter Stream ID field in order to associate the QUIC
+DATAGRAM frame with a stream. If an intermediary receives a QUIC DATAGRAM frame
+whose payload is too short to allow parsing the Quarter Stream ID field, the
+intermediary MUST treat it as an HTTP/3 connection error of type
+H3_GENERAL_PROTOCOL_ERROR. Intermediaries MUST ignore any HTTP/3 Datagram
+fields after the Quarter Stream ID.
+
+Endpoints parse both the Quarter Stream ID field and the Context ID field in
+order to associate the QUIC DATAGRAM frame with a stream and context within
+that stream. If an endpoint receives a QUIC DATAGRAM frame whose payload is too
+short to allow parsing the Quarter Stream ID field, the endpoint MUST treat it
+as an HTTP/3 connection error of type H3_GENERAL_PROTOCOL_ERROR. If an endpoint
+receives a QUIC DATAGRAM frame whose payload is long enough to allow parsing
+the Quarter Stream ID field but too short to allow parsing the Context ID
+field, the endpoint MUST abruptly terminate the corresponding stream with a
+stream error of type H3_GENERAL_PROTOCOL_ERROR.
 
 
 # CAPSULE HTTP/3 Frame Definition {#capsule-frame}
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -457,6 +457,160 @@ This registry is initially empty.
 
 --- back
 
+# Examples
+
+## CONNECT-UDP
+
+~~~
+Client                                             Server
+
+STREAM(44): HEADERS             -------->
+  :method = CONNECT-UDP
+  :scheme = https
+  :path = /
+  :authority = target.example.org:443
+
+STREAM(44): CAPSULE             -------->
+  Capsule Type = REGISTER_DATAGRAM_CONTEXT
+  Context ID = 0
+  Extension String = ""
+
+DATAGRAM                        -------->
+  Quarter Stream ID = 11
+  Context ID = 0
+  Payload = Encapsulated UDP Payload
+
+           <--------  STREAM(44): HEADERS
+                        :status = 200
+
+/* Wait for target server to respond to UDP packet. */
+
+           <--------  DATAGRAM
+                        Quarter Stream ID = 11
+                        Context ID = 0
+                        Payload = Encapsulated UDP Payload
+~~~
+
+
+## CONNECT-UDP with Timestamp Extension
+
+~~~
+Client                                             Server
+
+STREAM(44): HEADERS            -------->
+  :method = CONNECT-UDP
+  :scheme = https
+  :path = /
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-h3-datagram.md
```diff
@@ -117,28 +123,181 @@ following format (using the notation from the "Notational Conventions" section
...
 
 HTTP/3 Datagram Payload:
 
 : The payload of the datagram, whose semantics are defined by individual
 applications. Note that this field can be empty.
 
 Endpoints MUST treat receipt of a DATAGRAM frame whose payload is too short to
-parse the flow ID as an HTTP/3 connection error of type
+parse the Quarter Stream ID or Context ID as an HTTP/3 connection error of type
 H3_GENERAL_PROTOCOL_ERROR.
```
LPardue, line 154: This requirement seems awkard to implement as specified. I think some of that comes from mixing length checks and parsing. To elabortate:


* it states endpoints, but what about intermediaries (they need to parse just the Quarter Stream ID but should they forward a frame whose length is exactly the Quarter Stream ID?)
* for an endpoint parsing both fields it probably needs to be an 'and' 
* since the fields are varints its harder to check the DATAGRAM payload length can satisfy the field requirements. The obvious case is if payload length is 1 (two minimally encoded varints) but other combos make it harder.

The way that H3 specs the requirement isn't a perfect fit but I think the spirit is more closely align with what you're trying to say - https://tools.ietf.org/html/draft-ietf-quic-http-34#section-7.1

DavidSchinazi: Good point, one of those is hop-by-hop but the other is end-to-end. I've tweaked the requirements here to reflect that.

draft-ietf-masque-h3-datagram.md
```diff
@@ -117,26 +123,192 @@ following format (using the notation from the "Notational Conventions" section
...
+that stream. If an endpoint receives a QUIC DATAGRAM frame whose payload is too
+short to allow parsing the Quarter Stream ID field, the endpoint MUST treat it
+as an HTTP/3 connection error of type H3_GENERAL_PROTOCOL_ERROR. If an endpoint
+receives a QUIC DATAGRAM frame whose payload is long enough to allow parsing
+the Quarter Stream ID field but too short to allow parsing the Context ID
+field, the endpoint MUST abruptly terminate the corresponding stream with a
+stream error of type H3_GENERAL_PROTOCOL_ERROR.
+
+
+# CAPSULE HTTP/3 Frame Definition {#capsule-frame}
```
tfpauly, line 170: Bikeshed bikeshed bikeshed...

draft-ietf-masque-h3-datagram.md
```diff
@@ -117,26 +123,192 @@ following format (using the notation from the "Notational Conventions" section
...
+}
+~~~
+{: #close-capsule-format title="CLOSE_DATAGRAM_CONTEXT Capsule Format"}
+
+Context ID:
+
+: The context ID to close.
+
+Note that this close is unilateral and bidirectional: the sender of the frame
+unilateraly informs its peer of the closure. Endpoints can use
```
tfpauly, line 272:
```suggestion
unilaterally informs its peer of the closure. Endpoints can use
```

draft-ietf-masque-h3-datagram.md
```diff
@@ -117,26 +123,192 @@ following format (using the notation from the "Notational Conventions" section
...
+CLOSE_DATAGRAM_CONTEXT capsules to close a context that was initially
+registered by either themself, or by their peer. Endpoints MAY use the
```
tfpauly, line 274:
```suggestion
registered by either themselves, or by their peer. Endpoints MAY use the
```

tfpauly (approved):

draft-ietf-masque-h3-datagram.md
```diff
@@ -56,14 +54,8 @@ mechanism to send unreliable data while leveraging the security and
...
-HTTP/3 {{!H3=I-D.ietf-quic-http}}. It defines logical flows identified by a
-non-negative integer that are present at the start of the DATAGRAM frame
-payload. Flows are associated with HTTP messages using the Datagram-Flow-Id
-header field, allowing endpoints to match unreliable DATAGRAMS frames to the
-HTTP messages that they are related to.
-
-This design mimics the use of Stream Types in HTTP/3, which provide a
-demultiplexing identifier at the start of each unidirectional stream.
+HTTP/3 {{!H3=I-D.ietf-quic-http}}. It associates datagrams with client-initiated
+bidirectional streams and defines an optional additional demultiplexing layer.
```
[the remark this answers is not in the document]
DavidSchinazi: Indeed :)

draft-ietf-masque-h3-datagram.md
```diff
@@ -79,35 +71,49 @@ document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
...
+given HTTP request. Conceptually, the first layer of multiplexing is per-hop,
+while the second is end-to-end (see {{datagram-contexts}}).
+
 
-Flows are bidirectional exchanges of datagrams within a single QUIC connection.
-These are conceptually similar to streams in the sense that they allow
-multiplexing of application data. Flows are identified within a connection by a
-numeric value, referred to as the flow ID.  A flow ID is a 62-bit integer (0
-to 2^62-1).
+# Datagram Contexts {#datagram-contexts}
```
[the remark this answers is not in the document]
DavidSchinazi: Ah, that was my original intention. Thanks. Fixed.

draft-ietf-masque-h3-datagram.md
```diff
@@ -79,35 +71,49 @@ document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
...
 
-Flows lack any of the ordering or reliability guarantees of streams. Beyond
-this, a sender SHOULD ensure that DATAGRAM frames within a single flow are
-transmitted in order relative to one another.
+Contexts are birectional exchanges of datagrams associated with a given HTTP
```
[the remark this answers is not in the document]
DavidSchinazi: Done.

draft-ietf-masque-h3-datagram.md
```diff
@@ -117,26 +123,192 @@ following format (using the notation from the "Notational Conventions" section
...
 ~~~
 {: #h3-datagram-format title="HTTP/3 DATAGRAM Frame Format"}
 
-Flow ID:
+Quarter Stream ID:
+
+: A variable-length integer that contains the value of the client-initiated
+bidirectional stream that this datagram is associated with, divided by four.
+(The division by four stems from the fact that HTTP requests are sent on
+client-initiated bidirectional streams, and those have stream IDs equal to zero
```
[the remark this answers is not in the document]
DavidSchinazi: Reworded to use divisible instead of modulo

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Nope nope nope

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Done

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Done

DavidSchinazi (commented):
Thanks for the review!

draft-ietf-masque-h3-datagram.md
```diff
@@ -117,26 +123,192 @@ following format (using the notation from the "Notational Conventions" section
...
+{: #close-capsule-format title="CLOSE_DATAGRAM_CONTEXT Capsule Format"}
+
+Context ID:
+
+: The context ID to close.
+
+Note that this close is unilateral and bidirectional: the sender of the frame
+unilaterally informs its peer of the closure. Endpoints can use
+CLOSE_DATAGRAM_CONTEXT capsules to close a context that was initially
+registered by either themselves, or by their peer. Endpoints MAY use the
```
bemasc, line 274: Allowing endpoints to close contexts created by the peer seems confusing to me.  If my peer just closed the context I've been using ... what do I do?  Do I reopen it?  Do I fail the connection?

draft-ietf-masque-h3-datagram.md
```diff
@@ -117,26 +123,192 @@ following format (using the notation from the "Notational Conventions" section
...
+CLOSE_DATAGRAM_CONTEXT capsule to immediately reject a context that was just
```
bemasc, line 275: I think it would probably be better to distinguish CLOSE from "registration rejected".  Did the server close the context because it accepted my packets and believes the context is no longer needed?  Or did it close because I hit a resource limit?  Or was my request incompatible?

If something can be used to communicate failures, then I think it probably needs an error code, so the recipient knows what to do next.

draft-ietf-masque-h3-datagram.md
```diff
@@ -117,26 +123,192 @@ following format (using the notation from the "Notational Conventions" section
...
+as an HTTP/3 connection error of type H3_GENERAL_PROTOCOL_ERROR. If an endpoint
+receives a QUIC DATAGRAM frame whose payload is long enough to allow parsing
+the Quarter Stream ID field but too short to allow parsing the Context ID
+field, the endpoint MUST abruptly terminate the corresponding stream with a
+stream error of type H3_GENERAL_PROTOCOL_ERROR.
+
+
+# CAPSULE HTTP/3 Frame Definition {#capsule-frame}
+
+CAPSULE allows reliably sending request-related information end-to-end, even in
```
bemasc, line 172:
```suggestion
CAPSULE (placeholder, final name TBD) allows reliably sending request-related information end-to-end, even in
```

bemasc (approved):
I think this is a good change.  Thanks!

There are some things that I think we should still discuss, but we shouldn't hold up this change on those details.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: I'd say that depends on the application. In CONNECT-UDP if you close the main context then things break, so you might as well close the stream. But as another example, if your peer introduced a CONNECT-IP compression context and used it for a while and you're seeing low usage of it and want to reclaim memory, you can close than context and have the tunnel still be functional. Let's move this conversation to #43.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: That's an interesting idea. I've filed #53 to track it so we don't forget after this is merged.

draft-ietf-masque-h3-datagram.md
```diff
@@ -341,11 +457,167 @@ This registry is initially empty.
...
+
+STREAM(44): CAPSULE            -------->
+  Capsule Type = REGISTER_DATAGRAM_CONTEXT
+  Context ID = 2
+  Extension String = "timestamp"
+
+DATAGRAM                       -------->
+  Quarter Stream ID = 11
+  Context ID = 2
+  Payload = Encapsulated UDP Payload With TImeStamp
```
afrind, line 535: TImestamp -> Timestamp

draft-ietf-masque-h3-datagram.md
```diff
@@ -276,59 +363,88 @@ HTTP/3 datagrams enabled on this endpoint.
...
++---------------------------+-------+---------------+
 ~~~
 
 
-## Flow Parameters {#iana-params}
+## Context Extension Keys {#iana-keys}
 
-This document will request IANA to create an "HTTP Datagram Flow
-Parameters" registry. Registrations in this registry MUST
+This document will request IANA to create an "HTTP Datagram Context
```
afrind, line 434: I assume this is for the 'extension-string' parameter in Register Datagram Context.  The word key first appears here.

draft-ietf-masque-h3-datagram.md
```diff
@@ -117,26 +123,192 @@ following format (using the notation from the "Notational Conventions" section
...
+that stream. If an endpoint receives a QUIC DATAGRAM frame whose payload is too
+short to allow parsing the Quarter Stream ID field, the endpoint MUST treat it
+as an HTTP/3 connection error of type H3_GENERAL_PROTOCOL_ERROR. If an endpoint
+receives a QUIC DATAGRAM frame whose payload is long enough to allow parsing
+the Quarter Stream ID field but too short to allow parsing the Context ID
+field, the endpoint MUST abruptly terminate the corresponding stream with a
+stream error of type H3_GENERAL_PROTOCOL_ERROR.
+
+
+# CAPSULE HTTP/3 Frame Definition {#capsule-frame}
```
afrind, line 170: After reading this document, the main benefit I see to CAPSULE is that it makes it easier to deploy future end-to-end extension "frames" without having to modify intermediaries (because H3 drops unknown frames).  It may also make it easier to extend in client <--> server cases where the H3 stack itself is not easily modifiable (owned by a third-party, say).

Eventually it would probably warrant its own document and not be buried in DATAGRAM?

draft-ietf-masque-h3-datagram.md
```diff
@@ -79,35 +71,49 @@ document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
...
-These are conceptually similar to streams in the sense that they allow
-multiplexing of application data. Flows are identified within a connection by a
-numeric value, referred to as the flow ID.  A flow ID is a 62-bit integer (0
-to 2^62-1).
+## Datagram Contexts {#datagram-contexts}
 
-Flows lack any of the ordering or reliability guarantees of streams. Beyond
-this, a sender SHOULD ensure that DATAGRAM frames within a single flow are
-transmitted in order relative to one another.
+Contexts refer to bidirectional exchanges of datagrams associated with a single
```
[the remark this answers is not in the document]
DavidSchinazi: Added more text to explain contexts

draft-ietf-masque-h3-datagram.md
```diff
@@ -79,35 +71,49 @@ document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
...
+HTTP request. Contexts are identified within the scope of a given request by a
+numeric value, referred to as the context ID. A context ID is a 62-bit integer
+(0 to 2^62-1).
 
+While stream IDs are a per-hop concept, context IDs are an end-to-end concept.
+In other words, if a datagram travels through multiple intermediaries on its
```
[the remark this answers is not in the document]
DavidSchinazi: Done

draft-ietf-masque-h3-datagram.md
```diff
@@ -79,35 +71,49 @@ document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
...
 frame to the appropriate application context.
 
-Even-numbered flow IDs are client-initiated, while odd-numbered flow
-IDs are server-initiated. This means that an HTTP/3 client
-implementation of the flow ID allocation service MUST only provide
-even-numbered IDs, while a server implementation MUST only provide
-odd-numbered IDs. Note that, once allocated, any flow ID can be
-used by both client and server - only allocation carries separate namespaces to
-avoid requiring synchronization.
+Even-numbered context IDs are client-initiated, while odd-numbered context IDs
```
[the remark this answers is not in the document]
DavidSchinazi: The next sentence explains this further, I prefer initiated over defined.

draft-ietf-masque-h3-datagram.md
```diff
@@ -117,26 +123,192 @@ following format (using the notation from the "Notational Conventions" section
...
+
+: The payload of the datagram, whose semantics are defined by individual
+applications. Note that this field can be empty.
+
+Intermediaries parse the Quarter Stream ID field in order to associate the QUIC
+DATAGRAM frame with a stream. If an intermediary receives a QUIC DATAGRAM frame
+whose payload is too short to allow parsing the Quarter Stream ID field, the
+intermediary MUST treat it as an HTTP/3 connection error of type
+H3_GENERAL_PROTOCOL_ERROR. Intermediaries MUST ignore any HTTP/3 Datagram
+fields after the Quarter Stream ID.
```
afrind, line 157: A generic HTTP library capable of receiving and sending datagrams could be used to build either an intermediary or an endpoint.  WIth this specification, It would need two sets of APIs, one that had the context ID (for endpoints) and one without (for intermediaries).  If datagrams carry context IDs then intermediaries should be able to parse them.

draft-ietf-masque-h3-datagram.md
```diff
@@ -79,35 +71,49 @@ document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
...
-implementation of the flow ID allocation service MUST only provide
-even-numbered IDs, while a server implementation MUST only provide
-odd-numbered IDs. Note that, once allocated, any flow ID can be
-used by both client and server - only allocation carries separate namespaces to
-avoid requiring synchronization.
+Even-numbered context IDs are client-initiated, while odd-numbered context IDs
+are server-initiated. This means that an HTTP/3 client implementation of the
+context ID allocation service MUST only provide even-numbered IDs, while a
+server implementation MUST only provide odd-numbered IDs. Note that, once
+allocated, any context ID can be used by both client and server - only
```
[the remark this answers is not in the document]
DavidSchinazi: I find used to be clearer

draft-ietf-masque-h3-datagram.md
```diff
@@ -117,26 +123,192 @@ following format (using the notation from the "Notational Conventions" section
...
+
+: A variable-length integer indicating the context ID of the datagram (see
+{{datagram-contexts}}).
+
+HTTP/3 Datagram Payload:
+
+: The payload of the datagram, whose semantics are defined by individual
+applications. Note that this field can be empty.
+
+Intermediaries parse the Quarter Stream ID field in order to associate the QUIC
```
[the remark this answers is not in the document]
DavidSchinazi: Done

draft-ietf-masque-h3-datagram.md
DavidSchinazi: I agree with your benefit analysis. I'd prefer to avoid splitting the document further though. Let's wait and see if anyone wants to use CAPSULE but cannot implement DATAGRAM.

draft-ietf-masque-h3-datagram.md
```diff
@@ -117,26 +123,192 @@ following format (using the notation from the "Notational Conventions" section
...
+
+
+# CAPSULE HTTP/3 Frame Definition {#capsule-frame}
+
+CAPSULE allows reliably sending request-related information end-to-end, even in
+the presence of HTTP intermediaries.
+
+CAPSULE is an HTTP/3 Frame (as opposed to a QUIC frame) which SHALL only be sent
+in client-initiated bidirectional streams. Intermediaries MUST forward all
+received CAPSULE frame in their unmodified entirety on the same stream where it
```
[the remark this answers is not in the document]
DavidSchinazi: Done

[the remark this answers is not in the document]
DavidSchinazi: Capsules are by definition end-to-end. If intermediaries start looking inside, they might cause ossification. In theory, I'd happily use end-to-end encryption to protect them but we don't have keys for this purpose.

draft-ietf-masque-h3-datagram.md
```diff
@@ -117,26 +123,192 @@ following format (using the notation from the "Notational Conventions" section
...
+
+Endpoints which receive a Capsule with an unknown Capsule Type MUST silently
+drop that Capsule. Intermediaries MUST forward Capsules, even if they do not
+know the Capsule Type or cannot parse the Capsule Data.
+
+
+## The REGISTER_DATAGRAM_CONTEXT Capsule {#register-capsule}
+
+The REGISTER_DATAGRAM_CONTEXT capsule (type=0x00) allows an endpoint to inform
+its peer of the encoding and semantics of upcoming datagrams associated with
```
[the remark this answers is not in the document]
DavidSchinazi: Done

draft-ietf-masque-h3-datagram.md
```diff
@@ -117,26 +123,192 @@ following format (using the notation from the "Notational Conventions" section
...
+a given context ID. Its Capsule Data field consists of:
+
+~~~
+REGISTER_DATAGRAM_CONTEXT Capsule {
```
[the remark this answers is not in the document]
DavidSchinazi: I consider that to be a feature rather than a bug. If we don't exercise this path, implementors will cut corners and we won't be able to deploy extensions down the road. Requiring the main path to use the extension mechanism ensures that the mechanism is exercised and functional.

draft-ietf-masque-h3-datagram.md
```diff
@@ -117,26 +123,192 @@ following format (using the notation from the "Notational Conventions" section
...
+{{Section 3.2.6 of RFC7230}}):
+
+~~~
+  extension-string = [ ext-member *( "," ext-member ) ]
+  ext-member       = ext-member-key "=" ext-member-value
+  ext-member-key   = token
+  ext-member-value = token
+~~~
+
+Note that these registrations are unilateral and bidirectional: the sender of
```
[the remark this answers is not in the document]
DavidSchinazi: Done

draft-ietf-masque-h3-datagram.md
```diff
@@ -117,26 +123,192 @@ following format (using the notation from the "Notational Conventions" section
...
+the frame unilaterally defines the semantics it will apply to the datagrams it
+sends and receives using this context ID. Once a context ID is registered, it
+can be used in both directions.
+
+Endpoints MUST NOT send DATAGRAM frames using a Context ID until they have
+either sent or received a REGISTER_DATAGRAM_CONTEXT Capsule with the same
+Context ID. However, due to reordering, an endpoint that receives a DATAGRAM
+frame with an unknown Context ID MUST NOT treat it as an error, it SHALL
+instead drop the DATAGRAM frame silently, or buffer it for a short while in
```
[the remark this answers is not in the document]
DavidSchinazi: Done

draft-ietf-masque-h3-datagram.md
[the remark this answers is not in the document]
LPardue: I think adding a secure integrity mechanism is overkill. I'd treat these more like how headers and intermediaries work. Most intermediaries really have no business looking at the stuff, but some will decide to do stuff and if they break things then hopefully the failure to abide the spec will be reflected in market position.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: These are the keys from that section, I added references.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Done

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Your HTTP implementation is going to need to know whether to act on datagrams (endpoint) or forward them (intermediaries). Plumbing that bool to the parse function doesn't seem complex?

LPardue: What would intermediaries do with it? It might be fair to parse it and tell your API users "this field is entirely meaningless unless you are an endpoint".

DavidSchinazi (commented):
Thanks for the review!

draft-ietf-masque-h3-datagram.md
```diff
@@ -79,35 +71,49 @@ document are to be interpreted as described in BCP 14 {{!RFC2119}} {{!RFC8174}}
...
 frame to the appropriate application context.
 
-Even-numbered flow IDs are client-initiated, while odd-numbered flow
-IDs are server-initiated. This means that an HTTP/3 client
-implementation of the flow ID allocation service MUST only provide
-even-numbered IDs, while a server implementation MUST only provide
-odd-numbered IDs. Note that, once allocated, any flow ID can be
-used by both client and server - only allocation carries separate namespaces to
-avoid requiring synchronization.
+Even-numbered context IDs are client-initiated, while odd-numbered context IDs
```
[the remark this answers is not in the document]
DavidSchinazi: Odd-numbered context IDs allow servers to register their own contexts. For example, a CONNECT-IP server could choose to start compressing using a compression context of its choosing.

draft-ietf-masque-h3-datagram.md
afrind: > What would intermediaries do with it?

The same thing it does with other opaque things it get from the peer (eg: body bytes, datagram payloads) - forward to the other peer.

Whatever thing handles an HTTP stream would now have APIs like

```
onDatagram(payload)
sendDatagram(payload)
```
for intermediaries, or, if you are for endpoints:

```
onDatagram(contextID, payload)
sendDatagram(contextID, payload)
```

I'd prefer to have the library only have the second variant.  If using the library to implement a proxy, then the proxy implementation would pass the context ID and payload from one peer's `onDatagram` to the other side's `sendDatagram`. 

Logging/stats are another possible use.

I guess the broader point is -- as this spec is written, the datagram has to have that context ID, so why be so prescriptive about what the intermediary can and can't do with it.

draft-ietf-masque-h3-datagram.md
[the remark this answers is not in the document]
afrind: Most wg's approach grease by defining it and encouraging implementers to use it liberally enough to prevent this, but not require it.  HTTP/3 doesn't require a GREASE frame in order to function.

draft-ietf-masque-h3-datagram.md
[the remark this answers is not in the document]
LPardue: Yeah this makes sense, I just didn't digest all of the concepts before posting.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: We're considering making the Context ID optional (see #45). If intermediaries start parsing the context ID, it'll prevent us from doing that.

draft-ietf-masque-h3-datagram.md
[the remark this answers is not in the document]
DavidSchinazi: This isn't about grease, it's about using your extension mechanism. HTTP/3 forces all HTTP implementations to correctly support HTTP headers, because otherwise they couldn't send or parse `:method`. That guarantees that everyone correctly handles headers. We could have said that if no HEADERS frame is sent then it means GET, but we decided against that.

draft-ietf-masque-h3-datagram.md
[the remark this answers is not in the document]
afrind: I was being a bit tongue in cheek by suggesting security measures here. It strikes me as overkill too.

So is it explicitly forbidden for a proxy to register its own context ID on a stream via CAPSULE?  This would require some careful context-ID allocations by the proxy to avoid potential collisions, which in turn break the rules about the context-IDs being end-to-end.

I don't have a use case, but I'm not sure I see the the value in locking this down quite so tightly.  Are there designs that depend on the context identifier being the same at every hop?  Perhaps tunneling QUIC in 'forward' mode relies on this?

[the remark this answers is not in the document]
DavidSchinazi: The core tenet of the two-layer design is that Stream IDs are per-hop and context IDs are end-to-end. That allows extensions to rely on the fact that both endpoints have the same view of context IDs. CAPSULE exists as an end-to-end mechanism, and similarly relies on that property. If we need a new per-hop mechanism, then we can use a different h3 frame. I've clarified that intermediaries don't send capsules apart from forwarding them, so they're not allowed to inject their own context IDs.

draft-ietf-masque-h3-datagram.md
afrind: Whatever mechanism the endpoint will use to know if it should parse the context ID can also be used by the intermediary as well.  My point was that in this PR, they are required.

Presumably if/when negotiation is added the specification can be updated to reflect that.

DavidSchinazi: I've filed #54 to keep track of this discussion so it doesn't get lost when this is merged. Let's continue this conversation there.

DavidSchinazi:
@vasilvv you're referring to #45 which this PR isn't trying to address. That'll happen in a subsequent PR.

LPardue (approved):

PR #57, opened 2021-05-24
Clarify datagram lifetime with stream

### LINKED ISSUE #40: DATAGRAM and STREAM+FIN reordering
H3 datagrams are associated with a stream, so it makes sense that it is only valid to send DATAGRAMs before sending the stream FIN.  These packets may be received out of order however.  What should receivers do with these packets?

The simplest action for the receiver is to drop any datagrams received after FIN, but I can imagine applications where the receiver would prefer to allow them to be processed for some time.  Should the draft offer guidance for this?

The answer might be different if we use Flow IDs as connection scoped identifiers or the 2-layer approach (Stream ID+Flow ID) discussed at the interim.

DavidSchinazi:
Minor editorial clarification.

Closes #40.

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -182,11 +182,13 @@ the Quarter Stream ID field but too short to allow parsing the Context ID
 field, the endpoint MUST abruptly terminate the corresponding stream with a
 stream error of type H3_GENERAL_PROTOCOL_ERROR.
 
-If a DATAGRAM frame is received and its Quarter Stream ID maps to a stream that
-has already been closed, the receiver MUST silently drop that frame. If a
-DATAGRAM frame is received and its Quarter Stream ID maps to a stream that has
-not yet been created, the receiver SHALL either drop that frame silently or
-buffer it temporarily while awaiting the creation of the corresponding stream.
+Endpoints MUST NOT send HTTP/3 datagrams unless the corresponding stream is
+open. If an HTTP/3 datagram is received and its Quarter Stream ID maps to a
+stream that has already been closed, the receiver MUST silently drop that
+datagram. If an HTTP/3 datagram is received and its Quarter Stream ID maps to a
+stream that has not yet been created, the receiver SHALL either drop that
+datagram silently or buffer it temporarily while awaiting the creation of the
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-h3-datagram.md
```diff
@@ -182,11 +182,13 @@ the Quarter Stream ID field but too short to allow parsing the Context ID
...
-DATAGRAM frame is received and its Quarter Stream ID maps to a stream that has
-not yet been created, the receiver SHALL either drop that frame silently or
-buffer it temporarily while awaiting the creation of the corresponding stream.
+Endpoints MUST NOT send HTTP/3 datagrams unless the corresponding stream is
+open. If an HTTP/3 datagram is received and its Quarter Stream ID maps to a
+stream that has already been closed, the receiver MUST silently drop that
+datagram. If an HTTP/3 datagram is received and its Quarter Stream ID maps to a
+stream that has not yet been created, the receiver SHALL either drop that
+datagram silently or buffer it temporarily while awaiting the creation of the
+corresponding stream.
```
LPardue, line 191: There still seems to be some ambiguity here. Where you say "Endpoints MUST NOT send HTTP/3 datagrams unless the corresponding stream is open" do you really mean that it is ok to send Datagrams if the local send side is closed? CONNECT-UDP, IIRC, prohibits clients closing their end of the stream but there seems to be no requirement for other methods to do so. I think a model where you can only send DATAGRAM frames while you can send STREAM frames is the most rational way to approach this. 

Since there is no general ordering guarantee between DATAGRAM and STREAM, your recommendation "If an HTTP/3 datagram is received and its Quarter Stream ID maps to a stream that has already been closed, the receiver MUST silently drop" is good. We might want to loosen this a little and say "If an HTTP/3 receiver detects that it has received a datagram with a Quarter Stream ID that maps to a closed stream, it MUST silently drop." - the reason being that certain types of housekeeping might be impractical for some implementations.

DavidSchinazi: I tweaked the text to make clear it's tied to one direction of the stream, and that should resolve this.

LPardue (approved):
This improvement is good enough for me. Thanks.

PR #59, opened 2021-05-24
Clarify intermediary parsing rules

### LINKED ISSUE #54: Can intermediaries interact with capsules, contexts and context IDs?
In #52, capsules, contexts and context IDs are explicitly defined as end-to-end and there are requirements for intermediaries to not modify them, and to not even parse them. The motivation for this was to ensure that we retain the ability to deploy extensions end-to-end without having to modify intermediaries. @afrind mentioned that these requirements might be too strict, as he would like to have his intermediary be able to parse the context ID, for logging purposes.

DavidSchinazi:
Contexts are end-to-end, and intermediaries MUST NOT parse them.

Closes #54.

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -169,8 +169,12 @@ Intermediaries parse the Quarter Stream ID field in order to associate the QUIC
 DATAGRAM frame with a stream. If an intermediary receives a QUIC DATAGRAM frame
 whose payload is too short to allow parsing the Quarter Stream ID field, the
 intermediary MUST treat it as an HTTP/3 connection error of type
-H3_GENERAL_PROTOCOL_ERROR. Intermediaries MUST ignore any HTTP/3 Datagram
-fields after the Quarter Stream ID.
+H3_GENERAL_PROTOCOL_ERROR. Since the Context ID is optional and its use is
+negotiated end-to-end (and extensions MAY change this negotiation in a way
+which is opaque to intermediaries), intermediaries have no way of definitely
+knowing whether the Context ID field is present or absent. Because of this,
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-h3-datagram.md
```diff
@@ -169,8 +169,12 @@ Intermediaries parse the Quarter Stream ID field in order to associate the QUIC
...
 whose payload is too short to allow parsing the Quarter Stream ID field, the
 intermediary MUST treat it as an HTTP/3 connection error of type
-H3_GENERAL_PROTOCOL_ERROR. Intermediaries MUST ignore any HTTP/3 Datagram
-fields after the Quarter Stream ID.
+H3_GENERAL_PROTOCOL_ERROR. Since the Context ID is optional and its use is
+negotiated end-to-end (and extensions MAY change this negotiation in a way
+which is opaque to intermediaries), intermediaries have no way of definitely
+knowing whether the Context ID field is present or absent. Because of this,
+intermediaries MUST ignore any HTTP/3 Datagram fields after the Quarter Stream
+ID.
```
LPardue, line 177:
```suggestion
H3_GENERAL_PROTOCOL_ERROR. The Context ID field is optional and its use is
negotiated end-to-end see {{ref}}. Therefore intermediaries cannot
know whether the Context ID field is present or absent and they 
MUST ignore any HTTP/3 Datagram fields after the Quarter Stream ID.
```

Here's an alternative that's less wordy at risk of losing some points you were trying to make. 

I don't think putting/repeating the negotiation requirement right here is too great, so I've replace the parenthesised clause with a reference to wherever that requirement actually lives.

DavidSchinazi: Took that text and tweaked it a bit more.

LPardue (approved):
LGTM! :shipit:

PR #60, opened 2021-05-24
Clarify Opaque Capsules

DavidSchinazi:
@afrind points out that in order to correctly forward DATAGRAMs between transports where the QUIC DATAGRAM frame isn't always available, intermediaries need the ability to parse some capsules.

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -472,6 +472,12 @@ forwards them: in other words, an intermediary MAY send a DATAGRAM Capsule to
 forward an HTTP/3 Datagram which was received in a QUIC DATAGRAM frame, and
 vice versa.
 
+Note that while DATAGRAM capsules are sent on a stream, intermediaries can
+reencode HTTP/3 datagrams into QUIC DATAGRAM frames over the next hop, and
+those could be dropped. Because of this, applications have to always consider
+HTTP/3 datagrams to be unreliable, even if they were initially sent in a
+capsule.
+
 
 # The H3_DATAGRAM HTTP/3 SETTINGS Parameter {#setting}
 
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -200,10 +200,10 @@ the presence of HTTP intermediaries.
 
 CAPSULE is an HTTP/3 Frame (as opposed to a QUIC frame) which SHALL only be
 sent in client-initiated bidirectional streams. Intermediaries forward received
-CAPSULE frames in their unmodified entirety on the same stream where it would
-forward DATA frames. Each Capsule Type determines whether it is opaque or
-transparent to intermediaries: opaque capsules are forwarded unmodified while
-transparent ones can be parsed or added by intermediaries.
+CAPSULE frames on the same stream where it would forward DATA frames. Each
+Capsule Type determines whether it is opaque or transparent to intermediaries:
+opaque capsules are forwarded unmodified while transparent ones can be parsed,
+added, or removed by intermediaries.
 
 This specification of CAPSULE currently uses HTTP/3 frame type 0xffcab5. If this
 document is approved, a lower number will be requested from IANA.
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -245,6 +245,10 @@ the same order.
 Endpoints which receive a Capsule with an unknown Capsule Type MUST silently
 drop that Capsule.
 
+Receipt of a CAPSULE HTTP/3 Frame on a stream that is not a client-initiated
+bidirectional stream MUST be treated as a connection error of type
+H3_FRAME_UNEXPECTED.
+
 
 ## The REGISTER_DATAGRAM_CONTEXT Capsule {#register-capsule}
 
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -199,12 +199,9 @@ CAPSULE allows reliably sending request-related information end-to-end, even in
 the presence of HTTP intermediaries.
 
 CAPSULE is an HTTP/3 Frame (as opposed to a QUIC frame) which SHALL only be
-sent in client-initiated bidirectional streams. Intermediaries MUST forward all
-received CAPSULE frames in their unmodified entirety on the same stream where
-it would forward DATA frames. Intermediaries MUST NOT send any CAPSULE frames
-other than the ones it is forwarding. Intermediaries respect the order of
-CAPSULE frames: if an intermediary receives two CAPSULE frames in a given
-order, it MUST forward them in the same order.
+sent in client-initiated bidirectional streams. Intermediaries forward received
+CAPSULE frames in their unmodified entirety on the same stream where it would
+forward DATA frames.
 
 This specification of CAPSULE currently uses HTTP/3 frame type 0xffcab5. If this
 document is approved, a lower number will be requested from IANA.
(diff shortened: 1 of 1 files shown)


afrind:
CAPSULE was easier to understand when it was fully opaque.  When should I choose to define a semi-opaque CAPSULE vs defining a new frame?  Aren't both constructs effectively hop-by-hop?

DavidSchinazi:
@afrind you're the one who asked for this :-)

The benefit of a semi-opaque capsule is that it allows defining new capsules without having to update every single intermediary in the chain, because intermediaries forward all unknown capsules.

draft-ietf-masque-h3-datagram.md
```diff
@@ -230,9 +227,20 @@ Capsule Data:
...
+in their unmodified entirety. Intermediaries MUST NOT send any opaque CAPSULE
+frames other than the ones it is forwarding. All Capsule Types defined in this
+document are opaque, with the exception of the DATAGRAM Capsule, see
+{{datagram-capsule}}. Definitions of new Capsule Types MAY specify that the
+newly introduced type is not opaque. Intermediaries MUST treat unknown Capsule
+Types as opaque.
+
+Intermediaries respect the order of CAPSULE frames: if an intermediary receives
+two CAPSULE frames in a given order, it MUST forward them in the same order.
+
```
afrind, line 241: Order of transparent CAPSULEs (such as DATAGRAM) is no longer guaranteed, because an H3 -> H2 intermediary can have reordering effects.

DavidSchinazi: Good catch, fixed

draft-ietf-masque-h3-datagram.md
```diff
@@ -199,12 +199,9 @@ CAPSULE allows reliably sending request-related information end-to-end, even in
...
 CAPSULE is an HTTP/3 Frame (as opposed to a QUIC frame) which SHALL only be
-sent in client-initiated bidirectional streams. Intermediaries MUST forward all
-received CAPSULE frames in their unmodified entirety on the same stream where
-it would forward DATA frames. Intermediaries MUST NOT send any CAPSULE frames
-other than the ones it is forwarding. Intermediaries respect the order of
-CAPSULE frames: if an intermediary receives two CAPSULE frames in a given
-order, it MUST forward them in the same order.
+sent in client-initiated bidirectional streams. Intermediaries forward received
+CAPSULE frames in their unmodified entirety on the same stream where it would
+forward DATA frames.
```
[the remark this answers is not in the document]
DavidSchinazi: I like transparent. Updated the text.

Yes, datagrams are not reliable, even if sent by capsule.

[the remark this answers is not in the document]
afrind: > I like transparent

Heh, now that I read it I like it less :) 

It's good that transparent is the opposite of opaque, but as I read the text it seems strange that an intermediary is removing or inserting something that's called 'transparent'.  

Naming is hard though.  The property of the 'transparent' capsule that makes them different is that they are convertible to other protocol elements or extensions (I'm imagining what WebTransport streams might look like if conveyed by capsule).  But opaque vs convertible seems weird too.  I hereby resign from this bikeshed.

> Yes, datagrams are not reliable, even if sent by capsule.

Is that true?  Capsules are always sent on an HTTP stream which is reliable.  You mean that a receiver has the option to drop a DATAGRAM capsule?  I didn't text to this effect.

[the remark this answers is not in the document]
DavidSchinazi: I also don't much care for bikesheds. I'm going to keep "transparent" until someone proposes something better.

I've added text to spell out that yes: datagrams are unreliable, even when initially sent in capsules.

draft-ietf-masque-h3-datagram.md
```diff
@@ -230,9 +229,21 @@ Capsule Data:
...
+Types as opaque.
+
+Intermediaries respect the order of opaque CAPSULE frames: if an intermediary
+receives two opaque CAPSULE frames in a given order, it MUST forward them in
+the same order.
+
 Endpoints which receive a Capsule with an unknown Capsule Type MUST silently
-drop that Capsule. Intermediaries MUST forward Capsules, even if they do not
-know the Capsule Type or cannot parse the Capsule Data.
+drop that Capsule.
```
LPardue, line 246: This para seems a bit like a fragment, maybe you could just append the clause to the end of the sentence 2 paras up? I.e. "Intermediaries MUST treat unknown Capsule Types as opaque and MUST silently drop the Capsule."

draft-ietf-masque-h3-datagram.md
```diff
@@ -199,12 +199,11 @@ CAPSULE allows reliably sending request-related information end-to-end, even in
...
 
 CAPSULE is an HTTP/3 Frame (as opposed to a QUIC frame) which SHALL only be
-sent in client-initiated bidirectional streams. Intermediaries MUST forward all
-received CAPSULE frames in their unmodified entirety on the same stream where
-it would forward DATA frames. Intermediaries MUST NOT send any CAPSULE frames
-other than the ones it is forwarding. Intermediaries respect the order of
-CAPSULE frames: if an intermediary receives two CAPSULE frames in a given
-order, it MUST forward them in the same order.
+sent in client-initiated bidirectional streams. Intermediaries forward received
+CAPSULE frames in their unmodified entirety on the same stream where it would
```
afrind, line 203: Maybe just 'Intermediaries forward received CAPSULE frames on the same stream...' , since the capsule type (opacity) described in the next sentence explicitly carves out cases where capsules are not forwarded in their entirety.

draft-ietf-masque-h3-datagram.md
```diff
@@ -199,12 +199,11 @@ CAPSULE allows reliably sending request-related information end-to-end, even in
...
+forward DATA frames. Each Capsule Type determines whether it is opaque or
+transparent to intermediaries: opaque capsules are forwarded unmodified while
+transparent ones can be parsed or added by intermediaries.
```
afrind, line 206: Is 'parsed or added' what we want to say, or rather, 'transparent ones can be converted to semantically equivalent protocol features or extensions by an intermediary, as defined by the specific Capsule Type', or something like that.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: For intermediaries it's MUST forward, not MUST drop. Otherwise new capsules won't work without modifying intermediaries.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: I'd rather leave this vague. If it's transparent, the intermediary is allowed to do anything to a capsule if it wants to. I tweaked the text to clarify the intent.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: I agree, fixed.

draft-ietf-masque-h3-datagram.md
```diff
@@ -230,9 +229,21 @@ Capsule Data:
...
 : Data whose semantics depends on the Capsule Type.
 
+Unless otherwise specified, all Capsule Types are defined as opaque to
+intermediaries. Intermediaries MUST forward all received opaque CAPSULE frames
+in their unmodified entirety. Intermediaries MUST NOT send any opaque CAPSULE
+frames other than the ones it is forwarding. All Capsule Types defined in this
+document are opaque, with the exception of the DATAGRAM Capsule, see
+{{datagram-capsule}}. Definitions of new Capsule Types MAY specify that the
+newly introduced type is transparent. Intermediaries MUST treat unknown Capsule
+Types as opaque.
```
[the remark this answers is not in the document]
DavidSchinazi: I think this flows better here, see response on other comment

draft-ietf-masque-h3-datagram.md
LPardue: oh sorry I misread things

draft-ietf-masque-h3-datagram.md
```diff
@@ -199,12 +199,11 @@ CAPSULE allows reliably sending request-related information end-to-end, even in
 the presence of HTTP intermediaries.
 
 CAPSULE is an HTTP/3 Frame (as opposed to a QUIC frame) which SHALL only be
-sent in client-initiated bidirectional streams. Intermediaries MUST forward all
-received CAPSULE frames in their unmodified entirety on the same stream where
-it would forward DATA frames. Intermediaries MUST NOT send any CAPSULE frames
-other than the ones it is forwarding. Intermediaries respect the order of
-CAPSULE frames: if an intermediary receives two CAPSULE frames in a given
-order, it MUST forward them in the same order.
+sent in client-initiated bidirectional streams. Intermediaries forward received
```
LPardue, line 202: Not specific to this PR but the changes made me realise we probably want to specific what receivers of CAPSULE on the wrong type of stream do. I.E. if I understand CAPSULE and I see one arrive on the peer's control stream, I probably should be closing the connection with H3_FRAME_UNEXPECTED. Maybe we should spin this off to a new issue?

LPardue (approved):

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Fixed

PR #69, opened 2021-07-14
Do not convert datagrams to capsules

### LINKED ISSUE #65: MTU handling
This has been sent to the MASQUE WG mailing list in hopes to get some discussion there. However, I want an issue that tracks this: 

In the work of writing up our Connect-IP proposal (https://datatracker.ietf.org/doc/draft-kuehlewind-masque-connect-ip/) we looked into how to deal with MTU issues effectively. Our conclusion was that this is going to be a general problem for any user of HTTP Datagrams. Thus, we like to propose that MTU handling is done within HTTP datagram. 

This email will start explaining what we see as requirements for a MTU signaling solution for HTTP datagram, then propose a potential solution. 

Lets start with a figure that provides us with a framework to discuss the requirements:

```
+--------+ Path#1A +--------+        +-------+        +--------+
|Client A|<------->| HTTP   | Path#2 | MASQUE| Path#3 |        |
+--------+         | Inter- |<------>| Server|<------>| Target |
+--------+ Path#1B | mediary|        |(proxy)|        |        |
|Client B|<------->|        |        |       |        |        |
+--------+         +--------+        +-------+        +--------+
```

I think what makes this a bit more complex is the fact that we need to consider HTTP intermediaries, such as a front end load balancer that terminates a first QUIC and HTTP/3 connection between the client and that intermediary. From that intermediary another HTTP connection is used towards the HTTP server that consumes and produces the HTTP datagrams, and in the case of CONNECT-UDP and CONNECT-IP it also have a third path towards the target to consider. This figure includes two clients to remind us to consider that the HTTP intermediary may actually aggregate the HTTP request and response and HTTP datagram over one HTTP/2 or HTTP/3 connection over Path #2. In the case of MTU this complicates things more as the proxy cant assume that all HTTP requests have the same MTU for its datagrams on a HTTP/3 connection. And the Intermediary will be the entity that have direct knowledge of the client facing as well as the next hop MTU over the HTTP connections that may all differ in MTU. 

We also have to consider the fact that the underlying transport connection may at any time be subject to a IP MTU change due to route change for the path between the nodes. In addition if one have enabled PMTUD in TCP or QUIC a larger MTU on the individual path could be made available and in some case desirable to use. Thus, we need to consider dynamic changes during the HTTP connections life time and each HTTP request response pairs usage of HTTP datagram. 

So when using HTTP/3 datagram there are a strict MTU limit on the individual datagrams for it to be possible to be sent as QUIC datagrams, and not being forced to be encapsulated as CAPSULES over the reliable stream. This is clearly a possibility but results in that the datagrams are sent reliable and in order for each HTTP request, i.e. Connect-UDP or CONNECT-IP request. Also in case some end-to-end payloads fit in HTTP datagrams and others don’t there is potential for reordering among the payloads. Thus, to avoid this the client and the proxy needs to determine what the lowest currently supported HTTP datagram size on the path. 

For each QUIC connection the end point will know what the initial MTU value is for this path when the HTTP/3 connection has been successful completed. However, that knowledge will not be available if one attempt to construct and send the HTTP request prior to connection establishment has concluded.

So the requirements we see for an MTU handling solution for HTTP Datagrams are the following.

1)	Hop-by-hop signaling across the HTTP entities of the lowest MTU of any sub-path
2)	Needs to be associated to a particular HTTP request or end-to-end path to support aggregation by HTTP intermediaries
3)	Endpoints needs to be able to initiate update of the MTU value upon detection of any changes during HTTP Datagram streams lifetime.
4)	HTTP Intermediaries needs to be able to initiate updates of the MTU value upon detecting MTU changes from the individual HTTP connections.

Solution proposal
==============

A new HTTP Datagram Capsule is defined for MTU value exchange. This one is intended for the HTTP intermediary that needs to interpret, update or initiate sending of it. Thus, it needs to be a fixed registered type so it can be easily processed. It can also be exchanged in parallel to the Register_Datagram_* capsules as at that point the underlying HTTP connection is established and initial HTTP Datagram values will be known. A capsule will also travel all the way to the end. And an intermediary can initiate one in each direction for request paths. The only downside I see of this is that one is required per open stream when MTU changes occurs. Maybe someone have an idea of how to handle signaling when aggregating multiple endpoints streams onto one HTTP/3 connection. 

To make it more efficient, rather than sending one MTU capsule per stream, an MTU capsule could list all streams it is applicable to. That way the number of MTU capsules would be no more than the number of end-to-end paths actually used.

DavidSchinazi:
Credit to @bemasc for the proposal on the [mailing list](https://mailarchive.ietf.org/arch/msg/masque/7KJiS6bznHSlCqz1q8HTXoAk0to/).

Closes #65.

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -474,6 +474,12 @@ reencode HTTP Datagrams into QUIC DATAGRAM frames over the next hop, and those
 could be dropped. Because of this, applications have to always consider HTTP
 Datagrams to be unreliable, even if they were initially sent in a capsule.
 
+If an intermediary receives an HTTP Datagram in a QUIC DATAGRAM frame and is
+forwarding it on a connection that supports QUIC DATAGRAM frames, the
+intermediary SHOULD NOT convert that HTTP Datagram to a DATAGRAM capsule, as
+that would reduce the effectiveness of Datagram Packetization Layer Path MTU
+Discovery (DPLPMTUD) {{?RFC8899}}.
+
 
 # Context Extensibility {#context-ext}
 
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -476,9 +476,15 @@ Datagrams to be unreliable, even if they were initially sent in a capsule.
 
 If an intermediary receives an HTTP Datagram in a QUIC DATAGRAM frame and is
 forwarding it on a connection that supports QUIC DATAGRAM frames, the
-intermediary SHOULD NOT convert that HTTP Datagram to a DATAGRAM capsule, as
-that would reduce the effectiveness of Datagram Packetization Layer Path MTU
-Discovery (DPLPMTUD) {{?RFC8899}}.
+intermediary SHOULD NOT convert that HTTP Datagram to a DATAGRAM capsule. If
+the HTTP Datagram is too large to fit in a DATAGRAM frame because the MTU of
+that QUIC connection is too low, the intermediary SHOULD drop the HTTP Datagram
+instead of converting it to a DATAGRAM capsule. This preserves the end-to-end
+unreliability characteristic that methods such as Datagram Packetization Layer
+Path MTU Discovery (DPLPMTUD) depend on {{?RFC8899}}. An intermediary that
+converts QUIC DATAGRAM frames to DATAGRAM capsules allows HTTP Datagrams to be
+arbitrarily large without suffering any loss; this can misrepresent the true
+path properties, defeating methods such a DPLPMTUD.
 
 
 # Context Extensibility {#context-ext}
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -477,14 +477,16 @@ Datagrams to be unreliable, even if they were initially sent in a capsule.
 If an intermediary receives an HTTP Datagram in a QUIC DATAGRAM frame and is
 forwarding it on a connection that supports QUIC DATAGRAM frames, the
 intermediary SHOULD NOT convert that HTTP Datagram to a DATAGRAM capsule. If
-the HTTP Datagram is too large to fit in a DATAGRAM frame because the MTU of
-that QUIC connection is too low, the intermediary SHOULD drop the HTTP Datagram
-instead of converting it to a DATAGRAM capsule. This preserves the end-to-end
-unreliability characteristic that methods such as Datagram Packetization Layer
-Path MTU Discovery (DPLPMTUD) depend on {{?RFC8899}}. An intermediary that
-converts QUIC DATAGRAM frames to DATAGRAM capsules allows HTTP Datagrams to be
-arbitrarily large without suffering any loss; this can misrepresent the true
-path properties, defeating methods such a DPLPMTUD.
+the HTTP Datagram is too large to fit in a DATAGRAM frame (for example because
+the path MTU of that QUIC connection is too low or if the maximum UDP payload
+size advertised on that connection is too low), the intermediary SHOULD drop
+the HTTP Datagram instead of converting it to a DATAGRAM capsule. This
+preserves the end-to-end unreliability characteristic that methods such as
+Datagram Packetization Layer Path MTU Discovery (DPLPMTUD) depend on
+{{?RFC8899}}. An intermediary that converts QUIC DATAGRAM frames to DATAGRAM
+capsules allows HTTP Datagrams to be arbitrarily large without suffering any
+loss; this can misrepresent the true path properties, defeating methods such a
+DPLPMTUD.
 
 
 # Context Extensibility {#context-ext}


LPardue:
reading this in isolation, I'm a but unsure what the "effectiveness of DPLPMTUD" is supposed to mean.

Is there something implied here about how the intermediary will pack DATAGRAM frames into QUIC packets? For instance, is it being suggested that to make this work, we really need the intermediary not to pack out the size with other things?

DavidSchinazi:
The intent was to enable "end-to-end application DPLPMTUD" though I couldn't come up with a clear and concise name for it - in other words, there will be an application protocol that is running over HTTP Datagrams, and this is about the endpoints of that application performing DPLPMTUD at their layer. Does that make sense? If so, do you have thoughts on how best to word that?

bemasc:
> Is there something implied here about how the intermediary will pack DATAGRAM frames into QUIC packets?

The implication here is that the intermediary (1) knows a lower bound on the QUIC MTU and (2) never adds frames to a packet that is already larger than this bound (if it allows such packets at all).  I believe this is always true for QUIC, although I'm not sure it's stated explicitly anywhere.

DavidSchinazi:
@bemasc can you clarify what you meant by "a lower bound on the QUIC MTU" ? The MTU can change dynamically due to changes outside of QUIC's control so the only real lower bound is 1200 - and we need to be able to send packets larger than that.

bemasc:
I mean that, at any given time, a QUIC endpoint has a lower bound L on the estimated path MTU.  This lower bound is sometimes wrong, but it's never zero.  If an H3 DATAGRAM produces a QUIC packet that is larger than L, we just need to be sure that QUIC won't add other frames to this packet.  (I don't know if QUIC will send this packet or just drop the datagram, but either way, DPLPMTUD on the inner connection will work fine.)

LPardue:
The `SHOULD NOT` in the PR sounds fine. I'm still confused by the justification though, even after these comments.

```
           hop 1                     hop 2                           hop 3
Client  -->   Intermediary A  -->   Intermediary B  --> Target
```

For hop2, which is QUIC, it must have a minimum PMTU of 1200. The maximum PMTU could be bigger. The maximum QUIC packet size that can be sent should be the min(max MTU, max_udp_payload_size advertised by peer). After the intermediary gets the H3 DATAGRAM payload, it will need to reencode it. We're suggesting into a QUIC DATAGRAM frame, with H3 DATAGRAM headers. This reencoding could have different overhead sizes (hop2 might have larger QUIC packet overheads, or H3 DATAGRAM overheads.

Is the suggestion that the intermediary is using a PMTU size P1, and if it sees a H3 DATAGRAM that would produce a packet larger that P1, it should try to send it as a sort of probe?

bemasc:
> This reencoding could have different overhead sizes (hop2 might have larger QUIC packet overheads, or H3 DATAGRAM overheads.

That's fine, so long as the overhead of each hop rarely changes.

> Is the suggestion that the intermediary is using a PMTU size P1, and if it sees a H3 DATAGRAM that would produce a packet larger that P1, it should try to send it as a sort of probe?

It could try to send it, or it could drop it.  (It's not much of a "probe", since the intermediary won't necessarily know if it was dropped.)  Either way, the end-to-end DPLPMTUD will converge to some working value, which is all that matters.

LPardue:
So, can we remove this statement wrt to DPLPMTUD? I don't know what I'm supposed to apply it to.

DavidSchinazi:
I'm OK removing it, but in general I think it would be helpful to have the draft explain why that SHOULD NOT is there. And the motivation is to allow e2e DPLPMTUD. Because if the intermediary reencodes as capsules some of the time, the application-endpoints won't be able to get a sense for what packet sizes get through or not.

LPardue:
What you just said is much clearer than the proposed text. 

Longer term, I think what we'll probably end up with is a whole section that collects intermediary concerns. But until we arrive at that point, a better way to present the matter at hand could be something like the following, which is intended to follow on the theme of discussion from the paragraphs in that section.

"Intermediaries that receive HTTP Datagrams in QUIC DATAGRAM frames need to decide how to forward them. If the onward connection supports QUIC DATAGRAM frames, intermediaries SHOULD attempt to forward HTTP Datagrams in a QUIC frame. If the HTTP Datagram is too large to fit in a QUIC frame, the intermediary SHOULD drop it. This preserves the end-to-end unreliability characteristic that methods such as DPLPMTUD depend on. An intermediary that forwards HTTP Datagrams in DATAGRAM capsules allows HTTP Datagrams to be arbitrarily large. This can misrepresent the true path properties, defeating methods such a DPLPMTUD."

DavidSchinazi:
I like that, @LPardue. I've updated the text based on your comment, let me know what you think.

draft-ietf-masque-h3-datagram.md
```diff
@@ -474,6 +474,18 @@ reencode HTTP Datagrams into QUIC DATAGRAM frames over the next hop, and those
 could be dropped. Because of this, applications have to always consider HTTP
 Datagrams to be unreliable, even if they were initially sent in a capsule.
 
+If an intermediary receives an HTTP Datagram in a QUIC DATAGRAM frame and is
+forwarding it on a connection that supports QUIC DATAGRAM frames, the
+intermediary SHOULD NOT convert that HTTP Datagram to a DATAGRAM capsule. If
+the HTTP Datagram is too large to fit in a DATAGRAM frame because the MTU of
```
LPardue, line 480: Can we avoid using MTU here? https://www.rfc-editor.org/rfc/rfc9000.html#section-14 is relevant, unfortunately it calls it "Datagram size" which is a world of hurt. But the important thing is that the forwarding is limited by the the peer jut as much as the path. 

> Enforcement of the max_udp_payload_size transport parameter (Section 18.2) might act as an additional limit on the maximum datagram size.

Would "maximum UDP payload size (Section 18.2; RFC 9000)" work as a replacement.

DavidSchinazi: Can you clarify why would we want to avoid using MTU?

LPardue: Mainly because all of the hoops that RFC 9000 jumps through in order to spell out considerations for QUIC packet sizes beyond just MTU.

In the simplest sense, if an operator has configured a network in such a way to believes that MTU of the single hop is 1600 but the QUIC endpoints both advertise a max_udp_payload_size TP of 1400, they might be surprised when the larger size isn't used.

DavidSchinazi: That makes sense. How about phrasing it as examples (see latest commit)

LPardue (approved):

draft-ietf-masque-h3-datagram.md
LPardue: great, ship it!

gloinul:
So this is the first step to drop them. However, I am fairly convinced that we also need the signal to the Intermediaries when they need to send a HTTP datagram over the reliable stream instead of over the QUIC Datagram. Otherwise we will have MTU black holes for the minimal IPv6 packet size.

gloinul:
To expand on my previous comment. At least in CONNECT-IP and I think also for CONNECT-UDP. We do need a solution to ensure that we can tunnel and not blackhole inner payloads that correspond to the IPv6 minimal packet size. So for Connect-IP with all the overhead we need to send HTTP datagram containing a payload for 1280 IPv6 packets, otherwise we are not IPv6 compliant. For CONNECT-UDP this is not as clearly required as in the case of IPv6, however I think it would be a tradgic if this solution couldn't handle the required smallest allowed size of QUIC initial packets. 

That is why I think we need to have a solution where the endpoint can be certain that even if less efficient that the packets make it through if they are below these minimal sizes.

LPardue:
This sounds like a form of fragmentation to me. I'd rather fail hard than be fooled by unknown intermediaries that the path supports the size required. 

Instead of signalling the explicit MTUs as suggested on the mialing list, what about a capsule type equivalent to PTB that could be returned to the client to indicate something was too large and dropped. Then the client can make the active decision whether to reduce size (if possible) or retry by putting the HTTP Datagram in a DATAGRAM capsule itself.

gloinul:
@LPardue that I think works for me. In fact I think that is a better solution then the boundary value which requires state keeping and allows the HTTP endpoint to decided what to do. Also it can provide an endpoint with the actual current MTU. It is only executed when you can't forward the HTTP Datagram so it appear simpler than my original idea which would have needed cross layer interaction on changes in lower layers independent of when it occurs.

bemasc:
I think a PTB reply is likely unnecessary complexity.  Anything on CONNECT-IP has to exit the tunnel eventually, and it still needs PMTUD on that path.  ICMP connectivity is not reliable, so PLPMTUD is mandatory regardless, and will automatically handle any bottlenecks in the tunnel without the need for PTB.

LPardue:
just to clarify a little, I didn't mean a literal PTB. I meant to address Magnus' point that an end-to-end QUIC connection could fail to even get an client Initial through because the intermediary willingfully dropped it. A client might benefit from an immediate signal from the intermediary in such a case. I'm not sworn to this idea though.

bemasc:
Note that in CONNECT-UDP, the Initial should almost always travel in a CAPSULE, so it can be sent immediately ("false start") without worrying about reordering or packet loss.  That has ... interesting implications for MTU.

LPardue:
interesting, I hadn't thought of doing that. Right now my implementation buffers received datagrams, so whether I receive the DATAGRAM frame before a corresponding HEADERS is negligible. 

Agree it might have interesting implications.

gloinul:
@bemasc I don't think it is sufficient for us to simply declare that it is someone else problem with the MTU. From an Internet Architecture perspective the MASQUE tunnel is the equivalent of a link layer. There are certain requirements on link layers that carry IP packets, especially IPv6 packets. This requires support of the 1280 MTU, so we need a solution for that case. 

Secondly, I am a bit intrigued by your statement that in the case of Connect-UDP the first UDP packets would travel as capsule. I don't think the HTTP Datagram mandates that in any way. To my understanding, which appears to match @LPardue ones. Is that one can send a Connect-UDP request, followed by a HTTP Datagram context registration capsules for the needed Context IDs and then send HTTP Datagrams. There is nothing in the protocol that prevents this. Our Connect-UDP implementation will buffer these early HTTP datagrams and provide them to the Masque Server when it has processed the registration to create the IP/UDP packets based on the payload.

DavidSchinazi:
Hi folks, as per [our GitHub process](https://datatracker.ietf.org/doc/html/rfc8874#section-4.2.1) let's have this discussion somewhere that's not a GtiHub PR. I'd suggest [Magnus' email thread](https://mailarchive.ietf.org/arch/msg/masque/_fiWFDqS4UDKECR_cQW4bswhUkY/).

Since I haven't heard any objections on the text in this PR, I'm going to merge it. Let's discuss on the list whether we think more should be done about MTU or not.

PR #70, opened 2021-07-14
Specify how to support multiple drafts

DavidSchinazi:
This will help interop efforts before publication.

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -576,8 +576,8 @@ all cases, the maximum permitted value of the H3_DATAGRAM SETTINGS parameter is
 
 Some revisions of this draft specification use a different value for the
 H3_DATAGRAM Settings Parameter. This allows new draft revisions to make
-incompatible changes. Implementations MAY support multiple draft versions at
-the same time. In that scenario, endpoints MUST send multiple values for
+incompatible changes. Multiple draft versions MAY be supported by either
+endpoint in a connection. Such endpoints MUST send multiple values for
 H3_DATAGRAM. Once an endpoint has sent and received SETTINGS, it MUST compute
 the intersection of the values it has sent and received, and then it MUST
 select and use the most recent draft version from the intersection set. This
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -570,6 +570,20 @@ all cases, the maximum permitted value of the H3_DATAGRAM SETTINGS parameter is
 1.
 
 
+## Note About Draft Versions
+
+\[\[RFC editor: please remove this section before publication.]]
+
+Some revisions of this draft specification use a different value for the
+H3_DATAGRAM Settings Parameter. This allows new draft revisions to make
+incompatible changes. Implementations MAY support multiple draft versions at
+the same time. In that scenario, endpoints MUST send multiple values for
+H3_DATAGRAM. Once an endpoint has sent and received SETTINGS, it MUST compute
+the intersection of the values it has sent and received, and then it MUST
+select and use the most recent draft version from the intersection set. This
+ensures that both endpoints negotiate the same draft version.
+
+
 # Prioritization
 
 Prioritization of HTTP/3 datagrams is not defined in this document. Future


draft-ietf-masque-h3-datagram.md
```diff
@@ -570,6 +570,20 @@ all cases, the maximum permitted value of the H3_DATAGRAM SETTINGS parameter is
...
+\[\[RFC editor: please remove this section before publication.]]
+
+Some revisions of this draft specification use a different value for the
+H3_DATAGRAM Settings Parameter. This allows new draft revisions to make
+incompatible changes. Implementations MAY support multiple draft versions at
+the same time. In that scenario, endpoints MUST send multiple values for
+H3_DATAGRAM. Once an endpoint has sent and received SETTINGS, it MUST compute
+the intersection of the values it has sent and received, and then it MUST
+select and use the most recent draft version from the intersection set. This
+ensures that both endpoints negotiate the same draft version.
```
LPardue, line 584: This is a good change.

One thought, since `H3_DATAGRAM` is just a logical identifier we could have encoded the draft version into it and made it easier to talk about. But I think that ship has probably sailed.

DavidSchinazi: Agreed. That's what I plan on doing in our code. I think using `H3_DATAGRAM` in the draft and when it's ambiguous in email or code using `H3_DATAGRAM-03` seems fine.

draft-ietf-masque-h3-datagram.md
```diff
@@ -570,6 +570,20 @@ all cases, the maximum permitted value of the H3_DATAGRAM SETTINGS parameter is
...
 
+## Note About Draft Versions
+
+\[\[RFC editor: please remove this section before publication.]]
+
+Some revisions of this draft specification use a different value for the
+H3_DATAGRAM Settings Parameter. This allows new draft revisions to make
+incompatible changes. Implementations MAY support multiple draft versions at
+the same time. In that scenario, endpoints MUST send multiple values for
+H3_DATAGRAM. Once an endpoint has sent and received SETTINGS, it MUST compute
```
LPardue, line 581: this is a _little_ lose. I'm likely to write an implementation that supports both but would provide an option to let people run a client or server instance that supports a) supports multiple version b) only one version. This is important if I want to force testing of a newer version without falling back to an older one. 

For a suggested rewording:
```suggestion
incompatible changes. Multiple draft versions MAY be supported by either
endpoint in a connection. Such endpoints must MUST send multiple values for
H3_DATAGRAM. Once an endpoint has sent and received SETTINGS, it MUST compute
```

draft-ietf-masque-h3-datagram.md
```diff
@@ -570,6 +570,20 @@ all cases, the maximum permitted value of the H3_DATAGRAM SETTINGS parameter is
...
+the intersection of the values it has sent and received, and then it MUST
+select and use the most recent draft version from the intersection set. This
```
LPardue, line 583: Do we want to say anything about when the intersection is empty? Perhaps it's useful to close the connection with a specific error code like "H3_DATAGRAM_REQUIRED" or something. The receiving endpoint can then use that to see if it is due to mismatched sets, or that the peer was hoping for datagrams where they'll never be supported.

draft-ietf-masque-h3-datagram.md
LPardue: WFM

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Merged without the "must MUST"

draft-ietf-masque-h3-datagram.md
DavidSchinazi: I think that degrades easily to the case where the peer doesn't support datagram? Adding that error sounds reasonable, but is orthogonal to the multiple drafts question. Maybe file an issue and we can bring it up at 111?

LPardue: I'm happy to punt a special close code to a new issue but I think we need to fix this text up a bit more. The more i think about it, the more I'm unsure what the intent is.

Section 6 doesn't do a brilliant job of stating if both ends have to send a setting with value 1 in order for H3 Datagram to function or not. Theoretically, you could use true QUIC datagrams in one direction and datagram capsules in another (I'm not vouching to actually do that fwiw).

Extending to this PR, all one way to interpret this is that only the versions have to intersect. Not the values (0 or 1). It's implied that an endpoint advertises support with 1. But if I send you "old 0" and "new 1" and you send "old 1" then interop might fail because we agreed on a version but not how to use it.

DavidSchinazi: Discussed offline, let's merge this and have a follow up to clean up the exact semantics of the setting

PR #76, opened 2021-07-21
make CAPSULE frame ordering wrt HEADERS a global requirement

### LINKED ISSUE #73: Make Capsule Frame before Headers an H3_FRAME_UNEXPECTED connection error
The HTTP/3 RFC states 
> Receipt of an invalid sequence of frames MUST be treated 
> as a connection error of type H3_FRAME_UNEXPECTED;

The current draft instead recommends closing the stream rather than the connection and with the generic H3_GENERAL_PROTOCOL_ERROR

According to https://datatracker.ietf.org/doc/html/draft-ietf-quic-http-34#section-8.1 H3_GENERAL_PROTOCOL_ERROR is to be used when a more specific error code isn't available, but here we have clearly a more specific error available.

Also, what is the rationale for closing just the stream and not the whole connection?

Furthermore, a more general question:
right now the above recommendation only exists for REGISTER_DATAGRAM_NO_CONTEXT and REGISTER_DATAGRAM_CONTEXT capsule types. Would it be fair to say that no Capsule frame is allowed before headers instead?

LPardue:
This makes the requirement global to clients, servers, and
intermediaries.

closes #73

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -251,6 +251,12 @@ Receipt of a CAPSULE HTTP/3 Frame on a stream that is not a client-initiated
 bidirectional stream MUST be treated as a connection error of type
 H3_FRAME_UNEXPECTED.
 
+CAPSULE frames MUST NOT be sent on a stream before at least one HEADERS frame
+has been sent on that stream. This removes the need to buffer capsules when the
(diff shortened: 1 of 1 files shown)


DavidSchinazi (approved):

LPardue:
@lnicco does the PR address your issue?

LPardue:
merging, please reopen the ticket if you need anything more

PR #82, opened 2021-07-23
split the DATAGRAM capsule

### LINKED ISSUE #81: Consider splitting DATAGRAM Capsule into two types
The DATAGRAM capsule has an optional Context ID field. The presence of this field is based on state held in stream objects, not signalled on the wire itself. 

IME it's rare that the parsing of an HTTP/3 frame (in this case, the CAPSULE frame that holds the capsule) is dependent on state in this way.

So the suggestion is to define DATAGRAM/DATAGRAM_NO_CONTEXT capsule types which would allow stateless parsing of frames. After a whole frame is read, receivers can then validate ordering rules such REGISTER_CONTEXT followed by DATAGRAM capsule, REGISTER_NO_CONTEXT followed by DATAGRAM_NO_CONTEXT.

This would also allow an early shortcut exit where the CAPSULE frame length can be used to validate that the DATAGRAM capsule is suitable large.

LPardue:
Fixes #81.

This follows the explicit naming from REGISTER_CONTEXT and REGISTER_NO_CONTEXT but I've purposely defined both capsule types in the same section because the substantive considerations for both are mostly identical and I don't like duplication.

Previously we discussed the optionality of parsing the Context ID field. This change replaces that with requirements on endpoints to send the right datagram type, and reject the wrong datagram type, depending on the registration.

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -418,15 +418,19 @@ terminate the corresponding stream with a stream error of type
 H3_GENERAL_PROTOCOL_ERROR.
 
 
-## The DATAGRAM Capsule {#datagram-capsule}
+## The DATAGRAM and DATAGRAM_NO_CONTEXT Capsules {#datagram-capsule}
 
-The DATAGRAM capsule (type=0x02) allows an endpoint to send a datagram frame
-over an HTTP stream. This is particularly useful when using a version of HTTP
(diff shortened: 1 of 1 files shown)


LPardue:
I now realize the diff looks big but most of the later paragraphs are just substituting "DATAGRAM Capsule" and reflowing the text.

DavidSchinazi:
I'm not sure I agree with this direction, let's discuss on the issue (#81).

LPardue:
Ack

PR #87, opened 2021-09-09
Clarify that the H3_DATAGRAM setting applies only to datagrams

### LINKED ISSUE #85: Clarify meaning of SETTINGS parameter
In section 6, it says "A value of 0 indicates that this mechanism is not supported." The draft actually has two mechanisms: the DATAGRAM frame and the CAPSULE frame. It would be useful to clarify that the parameter only refers to support for the DATAGRAM frame.

DavidSchinazi:
As per h3 rules, it's fine to send capsules without a setting. The setting is about the semantics of datagrams. Also remove the requirement to send the datagram transport parameter because we now have capsules.

Closes #85

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -549,27 +549,20 @@ UTF-8.
 
 # The H3_DATAGRAM HTTP/3 SETTINGS Parameter {#setting}
 
-Implementations of HTTP/3 that support this mechanism can indicate that to
+Implementations of HTTP/3 that support HTTP Datagrams can indicate that to
 their peer by sending the H3_DATAGRAM SETTINGS parameter with a value of 1.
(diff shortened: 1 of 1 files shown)


LPardue:
I'm a bit confused about this one with the removal of the requirement. What purpose does the setting have now? Is it "I'm an endpoint that supports both H3 DATAGRAM frame and DATAGRAM capsule"? If so, using settings doesn't seem to gel well with the possible change of capsules to be conveyed in the datastream rather than in a CAPSULE frame.

LPardue:
Discussed offline and clarified. This spec defines the format of DATAGRAM frame or capsule.  The  SETTING helps articulate how the endpoint intends to consume things it receives. 

This might raise some interesting questions if someone tries to define a new formation the future. But they'll have to solve the problem they've made for themselves. This is fine as it is.

LPardue (approved):

PR #91, opened 2021-09-10
Rework how datagram format extensibility works

### LINKED ISSUE #66: How to actually use the Context ID registrations mechanism
Sent to the WG mailing list. 

I and my co-authors on https://datatracker.ietf.org/doc/draft-kuehlewind-masque-connect-ip/ have had some discussion around how one actually can and should be using the context-ID registration mechanism. This as we are actually need it to realize some of the functionality in our CONNECT-IP solution. 

We have found several ways that appear to practically work from an CONNECT-IP application perspective. 

A)	Statically define that in the context of CONNECT-IP the application will map certain formats to certain ContextID, and simply send registrations without any extension headers. 
B)	Define a Context Extension that is intended to provide format identifiers for the CONNECT-IP method different formats. 
C)	Use values or parameters in HTTP headers related to the format to carry the identifiers.
D)	Combine A and B, so that we have a set of initially statically defined mappings, and then use Context Extension for any future extensions. 

In our discussion we see that C appears to have some potential for failure modes if an HTTP intermediary does things with the headers. 

A) is very straight forward for an application initially but can run into trouble when needing to be extended but HTTP level negotiation of the general functionality should resolve those. 

B) appears to have some overhead from an application perspective as each application needs to define its own Context Extension to carry identifiers. Is it really intended that each HTTP application that needs to convey some identifier for its Context ID format to carry across the HTTP Datagram capsule. Would there be a point in defining a Context Extension that carries format identifier for what one like to use this context ID. 

D) appears to be simple initially, but leads to the extension mechanism having code paths that would not be exercised by default

So this was basically to start a discussion to see if we can get a bit more clearer guidance towards the HTTP application writers that will attempt to use HTTP Datagram on how to actually use Context IDs. I noted that the Connect-UDP method was no help as example yet.

### LINKED ISSUE #79: Walkthrough of intended use
The document currently has a lot of tools, but there are different ways of thinking about how you could use them and why they're in the protocol.  I feel like readability would improve dramatically if the document started with an overview that described one or two hypothetical flows.

Not fully fleshed out protocols -- just an example of declaring a context associated to a request, request headers that inform the context, context extensions that inform it, exchanging datagrams, and closing the context.  The document makes reference to a few possible applications; maybe one of these could be drawn a little bit larger?

DavidSchinazi:
After spending some time on CONNECT-IP, I realized that the current design won't work well. The primary motivation for including contexts and formatting in this document (as opposed to in the individual drafts) is that it allows to build a layer that can handle extensions without round trip negotiations: in particular it means clients can start using an extension format immediately without waiting for the HTTP response. For that to work, the server needs to drop unknown formats, so we need a single format enum as opposed to a list of them - that way the entire context is ignored instead of accepting the context but ignoring the extension that might have changed the encoding.

[Rendered view of this PR here](https://ietf-wg-masque.github.io/draft-ietf-masque-h3-datagram/rework_extensibility/draft-ietf-masque-h3-datagram.html).

Closes #66.
Closes #79.

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -103,6 +103,26 @@ hop, but the context ID will remain the same. Context IDs are opaque to
 intermediaries.
 
 
+## Datagram Formats {#datagram-formats}
+
+When an endpoint registers a datagram context (or the lack of contexts), it
+communicates the format (i.e., the semantics and encoding) of datagrams sent
+using this context. This is acccomplished by sending a Datagram Format Type as
+part of the registration capsule, see {{register-capsule}} and
+{{register-no-context-capsule}}. This type identifier is registered with IANA
+(see {{iana-format-types}}) and allows applications that use HTTP Datagrams to
+indicate what the content of datagrams are. Registration capsules carry a
+Datagram Format Data field which allows sending some additional information
+that would impact the format of datagrams.
+
+For example, a protocol which proxies IP packets can define a Datagram Format
+Type which represents an IP packet. The corresponding Datagram Format Data
+field would be empty. An extension to such a protocol that wishes to compress
+IP addresses could define a distinct Datagram Format Type and exchange two IP
+addresses via the Datagram Format Data field. Then any datagrams with that type
+would contain the IP packet with addresses elided.
+
+
 ## Context ID Allocation {#context-id-alloc}
 
 Implementations of HTTP Datagrams MUST provide a context ID allocation service.
@@ -320,7 +340,8 @@ field consists of:
 ~~~
 REGISTER_DATAGRAM_CONTEXT Capsule {
   Context ID (i),
-  Context Extensions (..),
+  Datagram Format Type (i),
+  Datagram Format Data (..),
 }
 ~~~
 {: #register-capsule-format title="REGISTER_DATAGRAM_CONTEXT Capsule Format"}
@@ -329,9 +350,16 @@ Context ID:
 
 : The context ID to register.
 
-Context Extensions:
+Datagram Format Type:
 
-: See {{context-ext}}.
+: A variable-length integer that defines the semantics and encoding of the HTTP
+Datagram Payload field of datagrams with this context ID, see
+{{datagram-formats}}.
+
+Datagram Format Data:
+
+: This field carries additional information that impact the format of datagrams
+with this context ID, see {{datagram-formats}}.
 
 Note that these registrations are unilateral and bidirectional: the sender of
 the frame unilaterally defines the semantics it will apply to the datagrams it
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-h3-datagram.md
```diff
@@ -103,6 +103,26 @@ hop, but the context ID will remain the same. Context IDs are opaque to
...
 
+## Datagram Formats {#datagram-formats}
+
+When an endpoint registers a datagram context (or the lack of contexts), it
+communicates the format (i.e., the semantics and encoding) of datagrams sent
+using this context. This is acccomplished by sending a Datagram Format Type as
+part of the registration capsule, see {{register-capsule}} and
+{{register-no-context-capsule}}. This type identifier is registered with IANA
+(see {{iana-format-types}}) and allows applications that use HTTP Datagrams to
+indicate what the content of datagrams are. Registration capsules carry a
```
LPardue, line 114: I read this sentence a few times and still how minor difficulty parsing it (in the context of the surrounding sentences). I don't have an immediate suggestion to improve readability though.

Looking at the example below, it seems this PR is proposing that the nature of DATAGRAM payload **always** has to be communicated by the type. That didn't used to be the case.

draft-ietf-masque-h3-datagram.md
```diff
@@ -103,6 +103,26 @@ hop, but the context ID will remain the same. Context IDs are opaque to
...
+Datagram Format Data field which allows sending some additional information
```
LPardue, line 115: Following from the above comment, since it sounds like datagrams always have a format, maybe it would be better to call this "Datagram Format Additional Data" to clearly mark it as separate to actual datagram data?

draft-ietf-masque-h3-datagram.md
DavidSchinazi: That's right, the idea would be to now always have an explicit type

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Good idea, done.

draft-ietf-masque-h3-datagram.md
```diff
@@ -466,11 +509,38 @@ endpoint receives a CLOSE_DATAGRAM_CONTEXT capsule that violates one or more of
...
+registered this context MUST NOT try to register another context with the same
+datagram format type on this stream.
+
+DENIED:
+
+: This indicates that the sender has rejected the context registration based on
+its local policy. The endpoint that had originally registered this context MUST
+NOT try to register another context with the same datagram format type and
+datagram format data on this stream.
+
```
LPardue, line 536: Another reason for rejecting a registration, for example, would be when receiving a Format Type of IP_PACKET on a CONNECT-UDP stream. Perhaps we can defined another code for that obviously broken case?

DavidSchinazi: Agreed, though the reaction to the code would be the same (MUST NOT use this type again) so I don't think we need a separate code

LPardue: yeah renaming DENIED to REJECTED and just stating policy or unsupported as examples could work. It's not a blocker to this PR though

LPardue (approved):

PR #97, opened 2021-09-27
expand capsule diagrams to include all fields

### LINKED ISSUE #94: Expand all capsule figures to include their Type and Length fields
QUIC and HTTP/3 style is to define the general format and then in each instance define the object including all of its fields. This would look like

```
REGISTER_DATAGRAM_CONTEXT Capsule {
  Type (i) = see Section 8.2,
  Length (i)
  Context ID (i),
  Datagram Format Type (i),
  Datagram Format Additional Data (..),
}
```

This can later be updated with the actual type value once things settle down

LPardue:
The thing that sucks here is that the diagram won't auto expand `{{iana-types}}`, so we can't easily ref the section. Manually referring to the section is a maintenance pain. So instead just state the actual type in the text an the diagram. I have a feeling you were linking to the section to avoid the maintenance hassle of updating type codes in multiple place. But I'm quite happy to take that burden because I find it more pleasant to read this way.

fixes #94

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -335,11 +335,12 @@ drop that Capsule.
 
 The REGISTER_DATAGRAM_CONTEXT capsule (see {{iana-types}} for the value of the
 capsule type) allows an endpoint to inform its peer of the encoding and
-semantics of datagrams associated with a given context ID. Its Capsule Data
-field consists of:
+semantics of datagrams associated with a given context ID.
 
 ~~~
 REGISTER_DATAGRAM_CONTEXT Capsule {
+  Type (i) = REGISTER_DATAGRAM_CONTEXT,
+  Length (i),
   Context ID (i),
   Datagram Format Type (i),
   Datagram Format Additional Data (..),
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-h3-datagram.md
```diff
@@ -333,13 +333,14 @@ drop that Capsule.
 
 ### The REGISTER_DATAGRAM_CONTEXT Capsule {#register-capsule}
 
-The REGISTER_DATAGRAM_CONTEXT capsule (see {{iana-types}} for the value of the
-capsule type) allows an endpoint to inform its peer of the encoding and
-semantics of datagrams associated with a given context ID. Its Capsule Data
-field consists of:
+The REGISTER_DATAGRAM_CONTEXT capsule (type=0xff37a1) allows an endpoint to
+inform its peer of the encoding and semantics of datagrams associated with a
+given context ID. Its Capsule Data field consists of:
```
DavidSchinazi, line 338: `Its Capsule Data field consists of` is no longer correct, we should rephrase it

DavidSchinazi (commented):
It's not just a maintenance hassle, I've been guilty of forgetting to update parts of redundant information like this in the past and I don't trust myself to not forget in the future - then you end up with a spec that disagrees with itself. How about instead:


```
The REGISTER_DATAGRAM_CONTEXT capsule (see {{iana-types}} for the value of the
capsule type) allows an endpoint to inform its peer of the encoding and
semantics of datagrams associated with a given context ID. Its Capsule Data
field consists of:

~~~
REGISTER_DATAGRAM_CONTEXT Capsule {
  Type (i) = REGISTER_DATAGRAM_CONTEXT,
  Length (i),
  Context ID (i),
  Datagram Format Type (i),
  Datagram Format Additional Data (..),
}
~~~
```

Also, to state the obvious: when we get to WGLC and don't expect to modify these numbers again, I fully support writing the final values into both the texts and diagrams to improve readability.

LPardue:
Perfectly reasonable

draft-ietf-masque-h3-datagram.md
LPardue: I vote for just remove it

DavidSchinazi: Fine by me

DavidSchinazi (approved):

PR #98, opened 2021-10-04
document the document structure

### LINKED ISSUE #92: Describe document structure and/or terms in intro section
There's now several different concepts to wrangle in this document. Reading from the top, we first tackle multiplexing (the major motivation for the document), then the frame, then capsules. Capsules underpin the mechanics of multiplexing. I don't think we can fix this chicken and egg problem but adding a brief overview of the concepts and terms before they are tackled in the prose could help readability.

LPardue:
fixes #92 

err'ing on the side of succinct


DavidSchinazi (approved):
Thanks for doing this!

PR #99, opened 2021-10-04
quarter stream ID has a legal range, state it and punishment for brea…

### LINKED ISSUE #93: Quarter Stream ID range is smaller than Stream ID range
Stating the obvious here. But we're using value with a legal range larger than semantically legal. RFC 9000 has some similar caveats with MAX_STREAMS https://www.rfc-editor.org/rfc/rfc9000.html#section-19.11

To address this, the suggestion is to add something like

>  The largest legal stream ID is 2^62 - 1, so this value cannot exceed (2^62-1) / 4. Receipt of a frame that includes a larger value MUST be treated as a connection error of type FRAME_ENCODING_ERROR

LPardue:
…king hte law

Also, switch to format to beatuiful superscript HTML rendering as used by RFC 9000.

fixes #93

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -161,11 +161,11 @@ HTTP/3 Datagram {
 Quarter Stream ID:
 
 : A variable-length integer that contains the value of the client-initiated
-bidirectional stream that this datagram is associated with, divided by four (The
+bidirectional stream that this datagram is associated with, divided by four (the
 division by four stems from the fact that HTTP requests are sent on
 client-initiated bidirectional streams, and those have stream IDs that are
 divisible by four.). The largest legal QUIC stream ID value is 2<sup>62</sup>-1,
-so the largest legal value value of Quarter Stream ID is 2<sup>62</sup>-1 / 4.
+so the largest legal value of Quarter Stream ID is 2<sup>62</sup>-1 / 4.
 Receipt of a frame that includes a larger value MUST be treated as a connection
 error of type FRAME_ENCODING_ERROR.
 
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -94,7 +94,7 @@ Contexts are optional, whether to use them or not is decided by the client on
 each request stream using registration capsules, see {{register-capsule}} and
 {{register-no-context-capsule}}. When contexts are used, they are identified
 within the scope of a given request by a numeric value, referred to as the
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-h3-datagram.md
```diff
@@ -161,10 +161,13 @@ HTTP/3 Datagram {
 Quarter Stream ID:
 
 : A variable-length integer that contains the value of the client-initiated
-bidirectional stream that this datagram is associated with, divided by four.
-(The division by four stems from the fact that HTTP requests are sent on
+bidirectional stream that this datagram is associated with, divided by four (The
```
DavidSchinazi, line 164: nit: if you remove the dot, then remove the following uppercase?
```suggestion
bidirectional stream that this datagram is associated with, divided by four (the
```

draft-ietf-masque-h3-datagram.md
```diff
@@ -161,10 +161,13 @@ HTTP/3 Datagram {
...
+division by four stems from the fact that HTTP requests are sent on
 client-initiated bidirectional streams, and those have stream IDs that are
-divisible by four.)
+divisible by four.). The largest legal QUIC stream ID value is 2<sup>62</sup>-1,
+so the largest legal value value of Quarter Stream ID is 2<sup>62</sup>-1 / 4.
```
DavidSchinazi, line 168: That's just too much value :P
```suggestion
so the largest legal value of Quarter Stream ID is 2<sup>62</sup>-1 / 4.
```

DavidSchinazi (approved):
Nice. Approved modulo a couple nits

draft-ietf-masque-h3-datagram.md
LPardue: derp

draft-ietf-masque-h3-datagram.md
```diff
@@ -161,10 +161,13 @@ HTTP/3 Datagram {
 Quarter Stream ID:
 
 : A variable-length integer that contains the value of the client-initiated
-bidirectional stream that this datagram is associated with, divided by four.
-(The division by four stems from the fact that HTTP requests are sent on
+bidirectional stream that this datagram is associated with, divided by four (the
+division by four stems from the fact that HTTP requests are sent on
 client-initiated bidirectional streams, and those have stream IDs that are
-divisible by four.)
+divisible by four.). The largest legal QUIC stream ID value is 2<sup>62</sup>-1,
```
DavidSchinazi, line 167: Well now we have an extra dot I think
```suggestion
divisible by four). The largest legal QUIC stream ID value is 2<sup>62</sup>-1,
```

PR #112, opened 2021-11-01
Treat malformed capsules as stream errors

afrind:
Borrowing text from H3.  Also I don't think there is a 'FRAME_ENCODING_ERROR' defined anywhere.

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -217,21 +217,20 @@ Intermediaries parse the Quarter Stream ID field in order to associate the QUIC
 DATAGRAM frame with a stream. If an intermediary receives a QUIC DATAGRAM frame
 whose payload is too short to allow parsing the Quarter Stream ID field, the
 intermediary MUST treat it as an HTTP/3 connection error of type
-H3_GENERAL_PROTOCOL_ERROR. The Context ID field is optional and whether it is
-present or not is decided end-to-end by the endpoints, see {{context-hdr}}.
-Therefore intermediaries cannot know whether the Context ID field is present or
-absent and they MUST ignore any HTTP/3 Datagram fields after the Quarter Stream
-ID.
+H3_DATAGRAM_ERROR. The Context ID field is optional and whether it is present or
+not is decided end-to-end by the endpoints, see {{context-hdr}}.  Therefore
+intermediaries cannot know whether the Context ID field is present or absent and
+they MUST ignore any HTTP/3 Datagram fields after the Quarter Stream ID.
 
 Endpoints parse both the Quarter Stream ID field and the Context ID field in
-order to associate the QUIC DATAGRAM frame with a stream and context within
-that stream. If an endpoint receives a QUIC DATAGRAM frame whose payload is too
-short to allow parsing the Quarter Stream ID field, the endpoint MUST treat it
-as an HTTP/3 connection error of type H3_GENERAL_PROTOCOL_ERROR. If an endpoint
-receives a QUIC DATAGRAM frame whose payload is long enough to allow parsing
-the Quarter Stream ID field but too short to allow parsing the Context ID
-field, the endpoint MUST abruptly terminate the corresponding stream with a
-stream error of type H3_GENERAL_PROTOCOL_ERROR.
+order to associate the QUIC DATAGRAM frame with a stream and context within that
+stream. If an endpoint receives a QUIC DATAGRAM frame whose payload is too short
+to allow parsing the Quarter Stream ID field, the endpoint MUST treat it as an
+HTTP/3 connection error of type H3_DATAGRAM_ERROR. If an endpoint receives a
+QUIC DATAGRAM frame whose payload is long enough to allow parsing the Quarter
+Stream ID field but too short to allow parsing the Context ID field, the
+endpoint MUST abruptly terminate the corresponding stream with a stream error of
+type H3_DATAGRAM_ERROR.
 
 Endpoints MUST NOT send HTTP/3 datagrams unless the corresponding stream's send
 side is open. On a given endpoint, once the receive side of a stream is closed,
@@ -364,7 +363,8 @@ the message as malformed or incomplete, according to the underlying transport
 protocol.  For HTTP/3, the handling of malformed messages is described in
 section 4.1.3 of {{H3}}.  For HTTP/2, the handling of malformed messages is
 described in section 8.1.2.6 of {{!H2=RFC7540}}.  For HTTP/1.1, the handling of
-incomplete messages is described in section 8 of {{!CORE-MESSAGING=I-D.draft-ietf-httpbis-messaging}}.
+incomplete messages is described in section 8 of
+{{!CORE-MESSAGING=I-D.draft-ietf-httpbis-messaging}}.
 
 Each capsule's payload MUST contain exactly the fields identified in its
 description. A capsule payload that contains additional bytes after the
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -361,7 +361,7 @@ drop that Capsule.
 
 Each capsule's payload MUST contain exactly the fields identified in its
 description. A capsule payload that contains additional bytes after the
-identified fields or a frame payload that terminates before the end of the
+identified fields or a capsule payload that terminates before the end of the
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-h3-datagram.md
```diff
@@ -196,7 +196,7 @@ client-initiated bidirectional streams, and those have stream IDs that are
 divisible by four). The largest legal QUIC stream ID value is 2<sup>62</sup>-1,
 so the largest legal value of Quarter Stream ID is 2<sup>62</sup>-1 / 4.
 Receipt of a frame that includes a larger value MUST be treated as a connection
-error of type FRAME_ENCODING_ERROR.
+error of type H3_FRAME_ERROR.
```
LPardue, line 199: This is a funny one.

HTTP/3 Datagram is not an HTTP/3 frame, so reporting this error code feels odd. 

I think it was me that picked FRAME_ENCODING_ERROR, which is defined in transport

> FRAME_ENCODING_ERROR (0x07):  An endpoint received a frame that was
      badly formatted -- for instance, a frame of an unknown type or an
      ACK frame that has more acknowledgment ranges than the remainder
      of the packet could carry.

But really, this is not a problem in the transport layer, and that also feels wrong now I think about it. I think the answer is a new application-layer `H3_DATAGRAM_ERROR`.

draft-ietf-masque-h3-datagram.md
```diff
@@ -357,6 +357,17 @@ the same order.
 Endpoints which receive a Capsule with an unknown Capsule Type MUST silently
 drop that Capsule.
 
+## Error Handling
+
+Each capsule's payload MUST contain exactly the fields identified in its
+description. A capsule payload that contains additional bytes after the
+identified fields or a frame payload that terminates before the end of the
```
LPardue, line 364:
```suggestion
identified fields or a capsule payload that terminates before the end of the
```

draft-ietf-masque-h3-datagram.md
afrind: I totally agree that borrowing error codes from lower level protocols is weird.  

H3_DATAGRAM_ERROR works when processing HTTP/3 DATAGRAMs.  But the capsule length errors don't fit with that.  Does Capsule Protocol needs its own error code range too (it can be conveyed over non-H3)?

draft-ietf-masque-h3-datagram.md
```diff
@@ -357,6 +357,17 @@ the same order.
...
+identified fields MUST be treated as a stream error of type H3_FRAME_ERROR. In
```
LPardue, line 365: as I said on the issue, this isn't an HTTP/3 frame error. Furthermore, we need to cover error handling for all HTTP versions; in H2 and H3 that's a stream error, in HTTP/1.1 it means tearing down the whole thing (which I think is acceptable because there is no way to recover the data stream) at this point.

draft-ietf-masque-h3-datagram.md
LPardue: I agree these are two different types of error

draft-ietf-masque-h3-datagram.md
```diff
@@ -196,7 +196,7 @@ client-initiated bidirectional streams, and those have stream IDs that are
 divisible by four). The largest legal QUIC stream ID value is 2<sup>62</sup>-1,
 so the largest legal value of Quarter Stream ID is 2<sup>62</sup>-1 / 4.
 Receipt of a frame that includes a larger value MUST be treated as a connection
-error of type FRAME_ENCODING_ERROR.
+error of type TBD.
```
DavidSchinazi, line 199: Can we define a new CAPSULE_ERROR? That'll be more helpful than TBD

draft-ietf-masque-h3-datagram.md
```diff
@@ -357,6 +357,22 @@ the same order.
...
+
+When an error occurs processing the capsule protocol, the receiver abruptly
+terminates the enclosing stream.  When running over HTTP/3 or HTTP/2, the
+receiver resets the stream.  When running over HTTP/1.1, the receiver terminates
+the underlying connection with an error.
+
+Each capsule's payload MUST contain exactly the fields identified in its
+description. A capsule payload that contains additional bytes after the
+identified fields or a capsule payload that terminates before the end of the
+identified fields MUST be treated as a stream error of type TBD. In
```
DavidSchinazi, line 370: Can we define a new CAPSULE_ERROR? That'll be more helpful than TBD

afrind: I thought about doing that here, but it seemed to violate the thesis.  You want it in this PR?

DavidSchinazi: What do you mean by "violate the thesis"? This is a stream error and not a connection error right?

afrind: Sorry, I meant "the thesis of this PR", which is about how to handle malformed capsules, not creating a new application error code space for capsule protocol.

DavidSchinazi: I didn't mean creating a new application error code space, just add a new value to the h3 errors registry

afrind: I don't think Capsule Protocol errors belong in the H3 space, since Capsule Protocol can be conveyed over other versions of HTTP, right?

DavidSchinazi: Capsules travel end-to-end, but HTTP aborts are hop-by-hop. Since we don't have a way to abort end-to-end, we're stuck with using an error from the corresponding HTTP version in use. But you're right, you'd need to allocate one for h2 as well, and mention that the stream abort mechanism in h1 is TCP RST.

afrind: > and mention that the stream abort mechanism in h1 is TCP RST.

Is this text sufficient?

>  When running over HTTP/1.1, the receiver terminates the underlying connection with an error.

I tried to use the language from core.

DavidSchinazi: SGTM. I just rather we either not specify the h3/h2 errors, or define them with a specific value as opposed to TBD.

LPardue: It isn't a problem to use H2/H3 error codes to communicate application (higher-layer) problems. I think it is useful to include the reason (bad capsule) but I could live with `H3_GENERAL_PROTOCOL_ERROR` or the H2 equivalent. 

If we consider a setup where there is an H3 -> H1.1 intermediary, the server behind the intermediary might detect a bad capsule and do a TCP RST. The intermediary can't know the reason for that and would probably end up having to use H3_GENERAL_PROTOCOL_ERROR. So the client might have to handle such a scenario anyway.

draft-ietf-masque-h3-datagram.md
```diff
@@ -196,7 +196,7 @@ client-initiated bidirectional streams, and those have stream IDs that are
 divisible by four). The largest legal QUIC stream ID value is 2<sup>62</sup>-1,
 so the largest legal value of Quarter Stream ID is 2<sup>62</sup>-1 / 4.
 Receipt of a frame that includes a larger value MUST be treated as a connection
-error of type TBD.
+error of type H3_DATAGRAM_ERROR.
```
DavidSchinazi, line 199: I think we'll also want to use H3_DATAGRAM_ERROR to replace any use of H3_GENERAL_PROTOCOL_ERROR in this section

draft-ietf-masque-h3-datagram.md
```diff
@@ -359,20 +359,21 @@ drop that Capsule.
...
-When an error occurs processing the capsule protocol, the receiver abruptly
-terminates the enclosing stream.  When running over HTTP/3 or HTTP/2, the
-receiver resets the stream.  When running over HTTP/1.1, the receiver terminates
-the underlying connection with an error.
+When an error occurs processing the capsule protocol, the receiver MUST treat
+the message as malformed or incomplete, according to the underlying transport
+protocol.  For HTTP/3, the handling of malformed messages is described in
+section 4.1.3 of {{H3}}.  For HTTP/2, the handling of malformed messages is
+described in section 8.1.2.6 of {{!H2=RFC7540}}.  For HTTP/1.1, the handling of
+incomplete messages is described in section 8 of {{!CORE-MESSAGING=I-D.draft-ietf-httpbis-messaging}}.
```
DavidSchinazi, line 367: nit: line wrap

DavidSchinazi (approved):
Thanks! Approved pending some minor tweaks

PR #119, opened 2021-11-09
Fixed the lack of parenthis resulting in that a non integer max value.

gloinul:


DavidSchinazi:
Why would parentheses be required? Exponent has higher precedence than subtraction.

LPardue:
I think this proposal resolves the interpretation of 2^62 - (1/4)

DavidSchinazi (approved):
Oh my bad, I had misread.

PR #122, opened 2021-11-09
Disallow capsule pipelining

### LINKED ISSUE #120: Serialization over HTTP/1.x of multiple HTTP requests with Datagram
In Section 2:

When running over HTTP/1, requests are strictly serialized in the connection, therefore the first layer of demultiplexing is not needed.

Is there any actual possibility to run more than a single request with HTTP datagram support over an transport connection? When one runs an HTTP request that enables datagram I don't see how it would be possible to close that request in any other way than closing the underlying transport connection. Or are there a way that I am missing that would enable one to close this request, and then issue another pipelined HTTP request that also uses HTTP datagrams? 

Is there a need to clarify this, or make it clear that only a single HTTP request with datagram can be sent?

DavidSchinazi:
Fixes #120

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -260,12 +260,12 @@ This specification defines the "data stream" of an HTTP request as the
 bidirectional stream of bytes that follow the headers in both directions. In
 HTTP/1.x, the data stream consists of all bytes on the connection that follow
 the blank line that concludes either the request header section, or the 2xx
-(Successful) response header section. (Note that HTTP pipelining is not
-supported with the capsule protocol.) In HTTP/2 and HTTP/3, the data stream of
-a given HTTP request consists of all bytes sent in DATA frames with the
-corresponding stream ID. The concept of a data stream is particularly relevant
-for methods such as CONNECT where there is no HTTP message content after the
-headers.
+(Successful) response header section. (Note that only a single HTTP request
+starting the capsule protocol can be sent on HTTP/1.x connections.) In HTTP/2
+and HTTP/3, the data stream of a given HTTP request consists of all bytes sent
+in DATA frames with the corresponding stream ID. The concept of a data stream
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-h3-datagram.md
```diff
@@ -260,7 +260,8 @@ This specification defines the "data stream" of an HTTP request as the
 bidirectional stream of bytes that follow the headers in both directions. In
 HTTP/1.x, the data stream consists of all bytes on the connection that follow
 the blank line that concludes either the request header section, or the 2xx
-(Successful) response header section. In HTTP/2 and HTTP/3, the data stream of
+(Successful) response header section. (Note that HTTP pipelining is not
```
LPardue, line 263: well it's not just pipelining, there's no means to send more than 1 request because the connection ceases to be HTTP after the request or response

DavidSchinazi: I was under the impression that "send more than 1 request" and "pipelining" were the exact same thing. Am I confused?

gloinul: I think the distinction between pipeling and non pipelining for a persistent connection is that you can send the additional HTTP request(s) after each other with pipelining without receiving any or the full response of the previous request. https://www.ietf.org/archive/id/draft-ietf-httpbis-messaging-19.html#name-pipelining

So, the restriction is really that it is not possible to subsequent HTTP requests after one that enables HTTP Datagram.

DavidSchinazi: Thanks, I've rewritten the text to replace pipelining with multiple requests

vasilvv: I feel like there should be some language in HTTP core that bans pipelining with `Connection: close` and `Connection: upgrade` that we would hopefully be covered by.

DavidSchinazi: @vasilvv I wasn't able to find that text, could you point me at it please?

Issue #111, opened 2021-10-27
RELIABLE_DATAGRAM


bemasc:
We should revisit [the RELIABLE_DATAGRAM question](https://mailarchive.ietf.org/arch/msg/masque/JImas3YR45txq60pZR2Ab-_LxzQ/).

martinthomson:
A discussion of this has broken out [here](https://github.com/ietf-wg-masque/draft-ietf-masque-h3-datagram/pull/124#discussion_r751879354).

My understanding is that there are two major motivating reasons for this:

1. The desire for end-to-end finite-sized message.  I think we resolved that capsules *of any type* provide this capability, so any attempt to add the capability to this draft is largely based on the desire to have a singular form.

2. Something about [zero-latency extensibility](https://github.com/ietf-wg-masque/draft-ietf-masque-h3-datagram/pull/124#discussion_r753423751).

The latter is the interesting one. I admit that I don't understand it well, especially the part that uses a RELIABLE_DATAGRAM capsule.

The desire to be able to negotiate different uses for QUIC DATAGRAM frames *and* for clients to be able to exercise those options prior to seeing any response from a server is very much context-specific.  Lowering latency is interesting, but I don't have a clear path from that requirement to this proposed mechanism.

Low-latency extensibility for DATAGRAM frames is something that can be enabled for each protocol by defining a "usage" field in the DATAGRAM payload for that protocol. The goal here seems to be to be able to impose the use of this "usage" field on any protocol, even if that protocol didn't originally choose to allow that.

That's quite a narrow applicability:

* uses DATAGRAM frames (unreliable, finite sized messages)
* has multiple uses for those DATAGRAM frames
* has some potential uses that are unknown at the time the protocol is defined (that naturally flows from the multiple uses scenario)
* wants clients to be able to send DATAGRAM frames immediately (which might well be a generic capability)
* didn't want the overhead of an additional field in DATAGRAM frames in the original design

It seems like this imposes a constraint on a client that wants to exercise this extensibility. Such a client can't send DATAGRAM frames immediately.  It has to use RELIABLE_DATAGRAM until the negotiation is complete.  That seems like an undesirable property to me, but I can see some advantages to it.  It's also unclear whether a server is similarly constrained.  It might answer a request without knowing that the client wants to support this extension, so it might also need to refrain from sending DATAGRAM frames until can perform the same negotiation.

[...]

PR #124, opened 2021-11-18
First PR for 2021 Design Team

DavidSchinazi:
This PR attempts to meet the majority of requirements from the design team's discussion on 2021-11-15, while removing all other features.

Note that this PR is against the design team consensus branch, not main.

[Rendered view of this PR](https://ietf-wg-masque.github.io/draft-ietf-masque-h3-datagram/design_team_output/draft-ietf-masque-h3-datagram.html).
[Rendered diff with the design team consensus branch](https://www.ietf.org/rfcdiff?url1=https://ietf-wg-masque.github.io/draft-ietf-masque-h3-datagram/draft-ietf-masque-h3-datagram.txt&url2=https://ietf-wg-masque.github.io/draft-ietf-masque-h3-datagram/design_team_output/draft-ietf-masque-h3-datagram.txt).

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -314,32 +314,30 @@ recent draft version from the intersection set. This ensures that both
 endpoints negotiate the same draft version.
 
 
-# The Sec-Capsule-Protocol Header
+# The Capsule-Protocol Header
 
 Definitions of new HTTP Methods or of new HTTP Upgrade Tokens that use the
-Capsule Protocol MAY use the Sec-Capsule-Protocol header field to simplify
+Capsule Protocol MAY use the Capsule-Protocol header field to simplify
 intermediary processing.
 
-"Sec-Capsule-Protocol" is an Item Structured Header {{!RFC8941}}. Its value MUST
-be a Boolean, its ABNF is:
+"Capsule-Protocol" is an Item Structured Header {{!RFC8941}}. Its value MUST be
+a Boolean, its ABNF is:
 
 ~~~
-Sec-Capsule-Protocol = sf-boolean
+Capsule-Protocol = sf-boolean
 ~~~
 
 Endpoints indicate that the Capsule Protocol is in use on the data stream by
-sending the Sec-Capsule-Protocol header field with a value of ?1. A
-Sec-Capsule-Protocol header field with a value of ?0 has the same semantics as
-when the header is not present. Intermediaries MAY use this header field to
-allow processing of HTTP Datagrams for unknown HTTP methods or unknown HTTP
-Upgrade Tokens.
+sending the Capsule-Protocol header field with a value of ?1. A Capsule-Protocol
+header field with a value of ?0 has the same semantics as when the header is not
+present. Intermediaries MAY use this header field to allow processing of HTTP
+Datagrams for unknown HTTP methods or unknown HTTP Upgrade Tokens.
 
-The Sec-Capsule-Protocol header field MUST NOT be sent multiple times on a
-message. The Sec-Capsule-Protocol header field MUST NOT be used on HTTP
-responses with a status code different from 2xx (Successful). This specification
-does not define any parameters for the Sec-Capsule-Protocol header field value,
-but future documents MAY define parameters. Receivers MUST ignore unknown
-parameters.
+The Capsule-Protocol header field MUST NOT be sent multiple times on a message.
+The Capsule-Protocol header field MUST NOT be used on HTTP responses with a
+status code different from 2xx (Successful). This specification does not define
+any parameters for the Capsule-Protocol header field value, but future documents
+MAY define parameters. Receivers MUST ignore unknown parameters.
 
 
 # Prioritization
@@ -382,7 +380,7 @@ Field Name" registry:
 
 Field Name:
 
-: Sec-Capsule-Protocol
+: Capsule-Protocol
 
 Template:
 
@@ -467,7 +465,7 @@ STREAM(44): HEADERS             -------->
   :scheme = https
   :path = /target.example.org/443/
   :authority = proxy.example.org:443
-  sec-capsule-protocol = ?1
+  capsule-protocol = ?1
 
 DATAGRAM                        -------->
   Quarter Stream ID = 11
@@ -475,7 +473,7 @@ DATAGRAM                        -------->
 
            <--------  STREAM(44): HEADERS
                         :status = 200
-                        sec-capsule-protocol = ?1
+                        capsule-protocol = ?1
 
 /* Wait for target server to respond to UDP packet. */
 
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -320,17 +320,19 @@ Definitions of new HTTP Methods or of new HTTP Upgrade Tokens that use the
 Capsule Protocol MAY use the Sec-Capsule-Protocol header field to simplify
 intermediary processing.
 
-Endpoints indicate that the Capsule Protocol is in use on the data stream by
-sending the Sec-Capsule-Protocol header field with a value of ?1. Intermediaries
-MAY use this header field to allow processing of HTTP Datagrams for unknown HTTP
-methods or unknown HTTP Upgrade Tokens.
-
 "Sec-Capsule-Protocol" is an Item Structured Header {{!RFC8941}}. Its value MUST
 be a Boolean, its ABNF is:
 
 ~~~
 Sec-Capsule-Protocol = sf-boolean
 ~~~
+
+Endpoints indicate that the Capsule Protocol is in use on the data stream by
+sending the Sec-Capsule-Protocol header field with a value of ?1. A
+Sec-Capsule-Protocol header field with a value of ?0 has the same semantics as
+when the header is not present. Intermediaries MAY use this header field to
+allow processing of HTTP Datagrams for unknown HTTP methods or unknown HTTP
+Upgrade Tokens.
 
 The Sec-Capsule-Protocol header field MUST NOT be sent multiple times on a
 message. The Sec-Capsule-Protocol header field MUST NOT be used on HTTP
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -321,9 +321,9 @@ Capsule Protocol MAY use the Sec-Capsule-Protocol header field to simplify
 intermediary processing.
 
 Endpoints indicate that the Capsule Protocol is in use on the data stream by
-sending the Sec-Capsule-Protocol header with a value of ?1. Intermediaries MAY
-use this header to allow processing of HTTP Datagrams for unknown HTTP methods
-or unknown HTTP Upgrade Tokens.
+sending the Sec-Capsule-Protocol header field with a value of ?1. Intermediaries
+MAY use this header field to allow processing of HTTP Datagrams for unknown HTTP
+methods or unknown HTTP Upgrade Tokens.
 
 "Sec-Capsule-Protocol" is an Item Structured Header {{!RFC8941}}. Its value MUST
 be a Boolean, its ABNF is:
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -229,24 +229,10 @@ Participant mode:
 : In this mode, the intermediary terminates the data stream and parses all
 Capsule Type and Capsule Length fields it receives.
 
-Each Capsule Type determines whether it is opaque or transparent to
-intermediaries in participant mode: opaque capsules are forwarded unmodified
-while transparent ones can be parsed, added, or removed by intermediaries.
-Intermediaries MAY modify the contents of the Capsule Data field of transparent
-capsule types.
-
-Unless otherwise specified, all Capsule Types are defined as opaque to
-intermediaries. Intermediaries MUST forward all received opaque CAPSULE frames
-in their unmodified entirety. Intermediaries MUST NOT send any opaque CAPSULE
-frames other than the ones it is forwarding. All Capsule Types defined in this
-document are opaque, with the exception of the DATAGRAM capsule, see
-{{datagram-capsule}}. Definitions of new Capsule Types MAY specify that the
-newly introduced type is transparent. Intermediaries MUST treat unknown Capsule
-Types as opaque.
-
-Intermediaries respect the order of opaque CAPSULE frames: if an intermediary
-receives two opaque CAPSULE frames in a given order, it MUST forward them in
-the same order.
+Capsules MUST be forwarded unmodified by intermediaries, with the exception of
+the DATAGRAM capsule; the DATAGRAM capsule MAY be parsed, added, or removed by
+intermediaries, see {{datagram-capsule}}. Definitions of new Capsule Types MAY
+specify custom intermediary processing.
 
 Endpoints which receive a Capsule with an unknown Capsule Type MUST silently
 drop that Capsule.
@@ -278,11 +264,11 @@ datagrams sent in QUIC DATAGRAM frames. In particular, the restrictions on when
 it is allowed to send an HTTP Datagram and how to process them from {{format}}
 also apply to HTTP Datagrams sent and received using the DATAGRAM capsule.
 
-The DATAGRAM capsule is transparent to intermediaries, meaning that
-intermediaries MAY parse them and send DATAGRAM capsules that they did not
-receive. This allows an intermediary to reencode HTTP Datagrams as it forwards
-them: in other words, an intermediary MAY send a DATAGRAM capsule to forward an
-HTTP Datagram which was received in a QUIC DATAGRAM frame, and vice versa.
+Intermediaries MAY parse DATAGRAM capsules and MAY send DATAGRAM capsules that
+they did not receive. This allows an intermediary to reencode HTTP Datagrams as
+it forwards them: in other words, an intermediary MAY send a DATAGRAM capsule to
+forward an HTTP Datagram which was received in a QUIC DATAGRAM frame, and vice
+versa.
 
 Note that while DATAGRAM capsules are sent on a stream, intermediaries can
 reencode DATAGRAM capsules into QUIC DATAGRAM frames over the next hop, and
@@ -307,10 +293,10 @@ DPLPMTUD.
 
 This document defines the RELIABLE_DATAGRAM capsule type (see {{iana-types}} for
 the value of the capsule type). This capsule has the same format and semantics
-as the DATAGRAM capsules, however it is opaque to intermediaries: intermediaries
-MUST NOT convert between RELIABLE_DATAGRAM caspules and other encodings of HTTP
-Datagrams, or vice versa. Endpoints MUST treat receipt of a RELIABLE_DATAGRAM
-capsule as equivalent to receipt of a DATAGRAM capsule with the same contents.
+as the DATAGRAM capsules, however intermediaries MUST NOT convert between
+RELIABLE_DATAGRAM caspules and other encodings of HTTP Datagrams, or vice versa.
+Endpoints MUST treat receipt of a RELIABLE_DATAGRAM capsule as equivalent to
+receipt of a DATAGRAM capsule with the same contents.
 
 
 # The H3_DATAGRAM HTTP/3 SETTINGS Parameter {#setting}
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -37,9 +37,9 @@ with a mechanism to send unreliable data while leveraging the security and
 congestion-control properties of QUIC. However, QUIC DATAGRAM frames do not
 provide a means to demultiplex application contexts. This document describes how
 to use QUIC DATAGRAM frames when the application protocol running over QUIC is
-HTTP/3. It associates datagrams with client-initiated bidirectional streams
-and defines an optional additional demultiplexing layer. Additionally, this
-document defines how to convey datagrams over prior versions of HTTP.
+HTTP/3. It associates datagrams with client-initiated bidirectional streams.
+Additionally, this document defines how to convey datagrams over prior versions
+of HTTP.
 
 
 --- middle
@@ -53,31 +53,20 @@ properties of QUIC. However, QUIC DATAGRAM frames do not provide a means to
 demultiplex application contexts. This document describes how to use QUIC
 DATAGRAM frames when the application protocol running over QUIC is HTTP/3
 {{!H3=I-D.ietf-quic-http}}. It associates datagrams with client-initiated
-bidirectional streams and defines an optional additional demultiplexing layer.
-Additionally, this document defines how to convey datagrams over prior versions
-of HTTP.
+bidirectional streams. Additionally, this document defines how to convey
+datagrams over prior versions of HTTP.
 
 This document is structured as follows:
 
 * {{multiplexing}} presents core concepts for multiplexing across HTTP versions.
-  * {{datagram-contexts}} defines datagram contexts, an optional end-to-end
-    multiplexing concept scoped to each HTTP request. Whether contexts are in
-    use is defined in {{context-hdr}}.
-  * {{datagram-formats}} defines datagram formats, which are scoped to contexts.
-    Formats communicate the format and encoding of datagrams sent using the
-    associated context.
-  * Contexts are identified using a variable-length integer. Requirements for
-    allocating identifier values are detailed in {{context-id-alloc}}.
 * {{format}} defines how QUIC DATAGRAM frames are used with HTTP/3. {{setting}}
   defines an HTTP/3 setting that endpoints can use to advertise support of the
   frame.
 * {{capsule}} introduces the Capsule Protocol and the "data stream" concept.
   Data streams are initiated using special-purpose HTTP requests, after which
   Capsules, an end-to-end message, can be sent.
-  * The following Capsule types are defined, together with guidance for defining new types:
-    * Datagram registration capsules {{register-capsule}}
-    * Datagram close capsule {{close-capsule}}
-    * Datagram capsules {{datagram-capsule}}
+  * {{datagram-capsule}} defines Datagram Capsule types, along with guidance
+    for specifying new capsule types.
 
 
 ## Conventions and Definitions {#defs}
@@ -91,85 +80,13 @@ when, and only when, they appear in all capitals, as shown here.
 # Multiplexing
 
 When running over HTTP/3, multiple exchanges of datagrams need the ability to
-coexist on a given QUIC connection. To allow this, HTTP datagrams contain two
-layers of multiplexing. First, the QUIC DATAGRAM frame payload starts with an
-encoded stream identifier that associates the datagram with a given QUIC
-stream. Second, datagrams optionally carry a context identifier (see
-{{datagram-contexts}}) that allows multiplexing multiple datagram contexts
-related to a given HTTP request. Conceptually, the first layer of multiplexing
-is per-hop, while the second is end-to-end.
+coexist on a given QUIC connection. To allow this, the QUIC DATAGRAM frame
+payload starts with an encoded stream identifier that associates the datagram
+with a given QUIC stream.
 
-When running over HTTP/2, the first level of demultiplexing is provided by the
-HTTP/2 framing layer. When running over HTTP/1, requests are strictly
-serialized in the connection, therefore the first layer of demultiplexing is
-not needed.
-
-
-## Datagram Contexts {#datagram-contexts}
-
-Within the scope of a given HTTP request, contexts provide an additional
-demultiplexing layer. Contexts determine the encoding of datagrams, and can be
-used to implicitly convey metadata. For example, contexts can be used for
-compression to elide some parts of the datagram: the context identifier then
-maps to a compression context that the receiver can use to reconstruct the
-elided data.
-
-While stream IDs are a per-hop concept, context IDs are an end-to-end concept.
-In other words, if a datagram travels through one or more intermediaries on its
-way from client to server, the stream ID will most likely change from hop to
-hop, but the context ID will remain the same. Context IDs are opaque to
-intermediaries.
-
-Contexts are OPTIONAL to implement for both endpoints. Intermediaries do not
-require any context-specific software to enable contexts. When contexts are
-supported by the implementation, their use is optional and can be selected on
-each stream. Endpoints inform their peer of whether they wish to use contexts
-via the Sec-Use-Datagram-Contexts HTTP header, see {{context-hdr}}.
-
-When contexts are used, they are identified within the scope of a given request
-by a numeric value, referred to as the context ID. A context ID is a 62-bit
-integer (0 to 2<sup>62</sup>-1).
-
-
-## Datagram Formats {#datagram-formats}
-
-When an endpoint registers a datagram context (or the lack of contexts), it
-communicates the format (i.e., the semantics and encoding) of datagrams sent
-using this context. This is acccomplished by sending a Datagram Format Type as
-part of the datagram registration capsule, see {{register-capsule}}. This type
-identifier is registered with IANA (see {{iana-format-types}}) and allows
-applications that use HTTP Datagrams to indicate what the content of datagrams
-are. Registration capsules carry a Datagram Format Additional Data field which
-allows sending some additional information that would impact the format of
-datagrams.
-
-For example, a protocol which proxies IP packets can define a Datagram Format
-Type which represents an IP packet. The corresponding Datagram Format
-Additional Data field would be empty. An extension to such a protocol that
-wishes to compress IP addresses could define a distinct Datagram Format Type
-and exchange two IP addresses via the Datagram Format Additional Data field.
-Then any datagrams with that type would contain the IP packet with addresses
-elided.
-
-
-## Context ID Allocation {#context-id-alloc}
-
-Implementations of HTTP Datagrams that support datagram contexts MUST provide
-a context ID allocation service. That service will allow applications
-co-located with HTTP to request a unique context ID that they can subsequently
-use for their own purposes. The HTTP implementation will then parse the context
-ID of incoming HTTP Datagrams and use it to deliver the frame to the
-appropriate application context.
-
-Even-numbered context IDs are client-initiated, while odd-numbered context IDs
-are server-initiated. This means that an HTTP client implementation of the
-context ID allocation service MUST only provide even-numbered IDs, while a
-server implementation MUST only provide odd-numbered IDs. Note that, once
-allocated, any context ID can be used by both client and server - only
-allocation carries separate namespaces to avoid requiring synchronization.
-Additionally, note that the context ID namespace is tied to a given HTTP
-request: it is possible for the same numeral context ID to be used
-simultaneously in distinct requests.
+When running over HTTP/2, demultiplexing is provided by the HTTP/2 framing
+layer. When running over HTTP/1, requests are strictly serialized in the
+connection, therefore demultiplexing is not needed.
 
 
 # HTTP/3 DATAGRAM Format {#format}
@@ -181,7 +98,6 @@ of {{QUIC}}):
 ~~~
 HTTP/3 Datagram {
   Quarter Stream ID (i),
-  [Context ID (i)],
   HTTP Datagram Payload (..),
 }
 ~~~
@@ -198,40 +114,14 @@ so the largest legal value of Quarter Stream ID is (2<sup>62</sup>-1) / 4.
 Receipt of a frame that includes a larger value MUST be treated as a connection
 error of type FRAME_ENCODING_ERROR.
 
-Context ID:
-
-: A variable-length integer indicating the context ID of the datagram (see
-{{datagram-contexts}}). Whether or not this field is present depends on whether
-datagram contexts are in use on this stream, see {{context-hdr}}. If this QUIC
-DATAGRAM frame is reordered and arrives before the receiver knows whether
-datagram contexts are in use on this stream, then the receiver cannot parse
-this datagram and the receiver MUST either drop that datagram silently or
-buffer it temporarily.
-
 HTTP Datagram Payload:
 
 : The payload of the datagram, whose semantics are defined by individual
 applications. Note that this field can be empty.
 
-Intermediaries parse the Quarter Stream ID field in order to associate the QUIC
-DATAGRAM frame with a stream. If an intermediary receives a QUIC DATAGRAM frame
-whose payload is too short to allow parsing the Quarter Stream ID field, the
-intermediary MUST treat it as an HTTP/3 connection error of type
-H3_GENERAL_PROTOCOL_ERROR. The Context ID field is optional and whether it is
-present or not is decided end-to-end by the endpoints, see {{context-hdr}}.
-Therefore intermediaries cannot know whether the Context ID field is present or
-absent and they MUST ignore any HTTP/3 Datagram fields after the Quarter Stream
-ID.
-
-Endpoints parse both the Quarter Stream ID field and the Context ID field in
-order to associate the QUIC DATAGRAM frame with a stream and context within
-that stream. If an endpoint receives a QUIC DATAGRAM frame whose payload is too
-short to allow parsing the Quarter Stream ID field, the endpoint MUST treat it
-as an HTTP/3 connection error of type H3_GENERAL_PROTOCOL_ERROR. If an endpoint
-receives a QUIC DATAGRAM frame whose payload is long enough to allow parsing
-the Quarter Stream ID field but too short to allow parsing the Context ID
-field, the endpoint MUST abruptly terminate the corresponding stream with a
-stream error of type H3_GENERAL_PROTOCOL_ERROR.
+Receipt of a QUIC DATAGRAM frame whose payload is too short to allow parsing the
+Quarter Stream ID field MUST be treated as an HTTP/3 connection error of type
+H3_GENERAL_PROTOCOL_ERROR.
 
 Endpoints MUST NOT send HTTP/3 datagrams unless the corresponding stream's send
 side is open. On a given endpoint, once the receive side of a stream is closed,
@@ -303,6 +193,10 @@ Capsule Value:
 : The payload of this capsule. Its semantics are determined by the value of the
 Capsule Type field.
 
+Note that use of the Capsule Protocol is not required to use HTTP Datagrams.
+Definitions of new HTTP Methods or of new HTTP Upgrade Tokens can use HTTP
+Datagrams with their own data stream protocol.
+
 
 ## Requirements
 
@@ -345,7 +239,7 @@ Unless otherwise specified, all Capsule Types are defined as opaque to
 intermediaries. Intermediaries MUST forward all received opaque CAPSULE frames
 in their unmodified entirety. Intermediaries MUST NOT send any opaque CAPSULE
 frames other than the ones it is forwarding. All Capsule Types defined in this
-document are opaque, with the exception of the datagram capsules, see
+document are opaque, with the exception of the DATAGRAM capsule, see
 {{datagram-capsule}}. Definitions of new Capsule Types MAY specify that the
 newly introduced type is transparent. Intermediaries MUST treat unknown Capsule
 Types as opaque.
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-h3-datagram.md
```diff
@@ -577,12 +297,22 @@ size advertised on that connection is too low), the intermediary SHOULD drop
...
 DPLPMTUD.
 
 
+## The RELIABLE_DATAGRAM Capsule
+
+This document defines the RELIABLE_DATAGRAM capsule type (see {{iana-types}} for
+the value of the capsule type). This capsule has the same format and semantics
+as the DATAGRAM capsules, however it is opaque to intermediaries: intermediaries
+MUST NOT convert between QUIC DATAGRAM frames and RELIABLE_DATAGRAM capsules or
+vice versa. Endpoints MUST treat receipt of a RELIABLE_DATAGRAM capsule as
```
LPardue, line 312: what about RELIABLE_DATAGRAM -> DATAGRAM capsule conversion?

Maybe simpler to just state that they are passed through and MUST NOT be converted to anything else.

draft-ietf-masque-h3-datagram.md
```diff
@@ -773,11 +458,8 @@ This registry initially contains the following entries:
 
 | Capsule Type                 |   Value   | Specification |
 |:-----------------------------|:----------|:--------------|
-| REGISTER_DATAGRAM_CONTEXT    | 0xff37a1  | This Document |
-| REGISTER_DATAGRAM            | 0xff37a2  | This Document |
-| CLOSE_DATAGRAM_CONTEXT       | 0xff37a3  | This Document |
-| DATAGRAM_WITH_CONTEXT        | 0xff37a4  | This Document |
-| DATAGRAM                     | 0xff37a5  | This Document |
+| DATAGRAM                     | 0xff37a0  | This Document |
+| RELIABLE_DATAGRAM            | 0xff37a5  | This Document |
```
LPardue, line 462: appreciate this is just the first proposal but the format changes and codepoint usage here seems breaking, sp probably requires a bump to the H3_DATAGRAM setting codepoint.

draft-ietf-masque-h3-datagram.md
```diff
@@ -877,11 +497,7 @@ STREAM(44): HEADERS             -------->
   :scheme = https
   :path = /target.example.org/443/
   :authority = proxy.example.org:443
-
-STREAM(44): DATA                -------->
-  Capsule Type = REGISTER_DATAGRAM
-  Datagram Format Type = UDP_PAYLOAD
-  Datagram Format Additional Data = ""
+  sec-capsule-protocol = ?1
```
LPardue, line 500: I note the response doesn't have the equivalent header. I assume that's intended but I'm unclear why based on the header's definition.

LPardue (commented):
Thanks @DavidSchinazi, I believe this reflects the direction from the meeting. Just a few questions/nits

draft-ietf-masque-h3-datagram.md
DavidSchinazi: No that was an oversight, fixed.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Interestingly, these changes don't break backwards compatibility because the format is the same in the absence of the Sec-Use-Datagram-Contexts header. So we don't need to change the setting or codepoints - which is nice because it doesn't break existing WebTransport implementations.

LPardue: Right but previously the intermediary could manipulate DATAGRAM capsules with type 0xff37a5, and now the text says they can't do that because the semantics of the RELIABLE_DATAGRAM are different. Is the assumption that no one implemented an intermediary that procesed DATAGRAM before this change?

LPardue: actually, that's a problem with capsule protocol versioning not the H3 SETTING but I think my point stands

DavidSchinazi: Ah right, yes the (unstated) assumption was that no one has yet implemented an intermediary yet so we don't need to maintain backwards compatibility yet.

afrind:
High level comment: 

Do we need to be as generic about transparent and opaque now?  Perhaps we can strictly state how an intermediary can convert between native DATAGRAMs and Datagram Capsules in both directions, and must forward any other capsules unmodified unless otherwise specified in a capsule's definition.

If we are baking the Datagram Capsule into The Capsule Protocol (TCP - sorry, couldn't resist) with a code point in this document, then is it safe to say that the intended usage is for conveying datagrams and optional request related information over a request stream?  Eg: can another document specify the use "Capsule Protocol" but forbid the transmission of Datagram capsules, or does it become a different protocol at that point?   Note that WebTransport over H3 may do exactly this, since that protocol requires QUIC DATAGRAMs.  Can an application using WebTransport specify the use of Capsule Protocol for a non-request stream?

LPardue:
> Eg: can another document specify the use "Capsule Protocol" but forbid the transmission of Datagram capsules, or does it become a different protocol at that point? Note that WebTransport over H3 may do exactly this, since that protocol requires QUIC DATAGRAMs. 

IMO the use (or prohibition) of capsule types would be a job of a spec on top of this. The capsule protocol is still in effect even if you choose to allow only some types of capsules on it. The PR states "Note that use of the Capsule Protocol is not required to use HTTP Datagrams." so your WT use case seems to be supported.

> Can an application using WebTransport specify the use of Capsule Protocol for a non-request stream?

Interesting, that might be a case where we can state something like "a future extension to the capsule protocol could describe how it can be applied to other stream types beyond client-initiated bidirectional streams"

DavidSchinazi:
> Do we need to be as generic about transparent and opaque now? Perhaps we can strictly state how an intermediary can convert between native DATAGRAMs and Datagram Capsules in both directions, and must forward any other capsules unmodified unless otherwise specified in a capsule's definition.

I agree, I've removed the concept of opaque vs transparent.

> If we are baking the Datagram Capsule into The Capsule Protocol (TCP - sorry, couldn't resist) with a code point in this document, then is it safe to say that the intended usage is for conveying datagrams and optional request related information over a request stream? Eg: can another document specify the use "Capsule Protocol" but forbid the transmission of Datagram capsules, or does it become a different protocol at that point? Note that WebTransport over H3 may do exactly this, since that protocol requires QUIC DATAGRAMs.

I don't think that's right: WebTransport currently requires the QUIC DATAGRAM frame to be supported but it doesn't require datagrams to be sent using it. I think that's a fine place to be. As Lucas points out, WebTransport could decide to ban sending using DATAGRAM capsules while still using the capsule protocol.

> Can an application using WebTransport specify the use of Capsule Protocol for a non-request stream?

The capsule protocol is currently 1:1 tied to a request stream, so as specified I'd say it can't be used on WebTransport streams other than the WebTransport CONNECT stream. I'm not sure there's a use case for that though.

afrind:
> The capsule protocol is currently 1:1 tied to a request stream

I think we can make that even more explicit in the Section 4 intro.

> I'm not sure there's a use case for that though

There's a use case for framing on any unframed byte-stream, which WebTransport streams are.  Capsule Protocol minus Datagram Capsule could serve this purpose.  I'm ok saying that's out of scope though.

DavidSchinazi:
> > The capsule protocol is currently 1:1 tied to a request stream
> 
> I think we can make that even more explicit in the Section 4 intro.

Agreed, added some text.

> > I'm not sure there's a use case for that though
> 
> There's a use case for framing on any unframed byte-stream, which WebTransport streams are. Capsule Protocol minus Datagram Capsule could serve this purpose. I'm ok saying that's out of scope though.

I'd say that's out of scope. Applications can decide to use "TLVs where T and L are varints" without saying it's the capsule protocol.

draft-ietf-masque-h3-datagram.md
```diff
@@ -91,85 +80,13 @@ when, and only when, they appear in all capitals, as shown here.
...
-context ID allocation service MUST only provide even-numbered IDs, while a
-server implementation MUST only provide odd-numbered IDs. Note that, once
-allocated, any context ID can be used by both client and server - only
-allocation carries separate namespaces to avoid requiring synchronization.
-Additionally, note that the context ID namespace is tied to a given HTTP
-request: it is possible for the same numeral context ID to be used
-simultaneously in distinct requests.
+coexist on a given QUIC connection. To allow this, the QUIC DATAGRAM frame
+payload starts with an encoded stream identifier that associates the datagram
+with a given QUIC stream.
```
martinthomson, line 85:
```suggestion
payload starts with an encoded stream identifier that associates the datagram
with a request stream.
```

You already say "encoded stream identifier" and the goal is to associate with request streams.  In WebTransport, we might use those bidirectional streams without making them requests, which will require (or at least allow) some additional checking to see that the identified stream contains a request, but that's something WebTransport can handle.

draft-ietf-masque-h3-datagram.md
```diff
@@ -198,40 +114,14 @@ so the largest legal value of Quarter Stream ID is (2<sup>62</sup>-1) / 4.
 Receipt of a frame that includes a larger value MUST be treated as a connection
```
martinthomson, line 114: Noticed in reviewing the whole...  The line above says `(2<sup>62</sup>-1) / 4` where `2<sup>60</sup>-1` would do.

draft-ietf-masque-h3-datagram.md
```diff
@@ -250,8 +140,8 @@ corresponding stream.
 
 This specification introduces the Capsule Protocol. The Capsule Protocol is a
 sequence of type-length-value tuples that allows endpoints to reliably
-communicate request-related information end-to-end, even in the presence of
-HTTP intermediaries.
+communicate request-related information end-to-end on HTTP request streams, even
+in the presence of HTTP intermediaries.
 
 
 ## Capsule Protocol {#capsule-protocol}
```
martinthomson, line 147: There is text missing from the introduction here that establishes the conditions on the use of capsule protocol.  That is, it only applies to those protocols that use extended CONNECT and also opt in to the use of the capsule protocol.

The nice thing with the draft up until this point is that it doesn't constrain the use of datagrams to those protocols that use extended CONNECT.  This part does.

(I know that this is perhaps fanciful, but I can imagine uses for regular HTTP requests that are enhanced by the use of DATAGRAM frames.)

draft-ietf-masque-h3-datagram.md
```diff
@@ -303,6 +193,10 @@ Capsule Value:
 : The payload of this capsule. Its semantics are determined by the value of the
 Capsule Type field.
 
+Note that use of the Capsule Protocol is not required to use HTTP Datagrams.
+Definitions of new HTTP Methods or of new HTTP Upgrade Tokens can use HTTP
+Datagrams with their own data stream protocol.
+
```
martinthomson, line 199: This bit belongs up front.

draft-ietf-masque-h3-datagram.md
```diff
@@ -335,238 +229,50 @@ Participant mode:
...
-newly introduced type is transparent. Intermediaries MUST treat unknown Capsule
-Types as opaque.
-
-Intermediaries respect the order of opaque CAPSULE frames: if an intermediary
-receives two opaque CAPSULE frames in a given order, it MUST forward them in
-the same order.
+Capsules MUST be forwarded unmodified by intermediaries, with the exception of
+the DATAGRAM capsule; the DATAGRAM capsule MAY be parsed, added, or removed by
+intermediaries, see {{datagram-capsule}}. Definitions of new Capsule Types MAY
+specify custom intermediary processing.
```
martinthomson, line 235: 👍

I might amend this to include the condition that the intermediary understand the semantics of the request enough to know that capsules are in use:

```suggestion
Capsules MUST be forwarded unmodified by intermediaries, with the exception of
the DATAGRAM capsule. An intermediary that understands the request semantics 
sufficient to know that capsules are in use MAY parse, add, or remove DATAGRAM
capsules; see {{datagram-capsule}}. Definitions of new Capsule Types MAY specify
custom intermediary processing.
```

What I don't want to see is intermediaries guessing that a particular unknown `:protocol` is using capsules when they do not.  Splicing in DATAGRAM frames/capsules in that context might go badly wrong.

draft-ietf-masque-h3-datagram.md
```diff
@@ -335,238 +229,50 @@ Participant mode:
 : In this mode, the intermediary terminates the data stream and parses all
 Capsule Type and Capsule Length fields it receives.
```
martinthomson, line 230: I think that the text on the two intermediary processing modes can be removed.

draft-ietf-masque-h3-datagram.md
```diff
@@ -335,238 +229,50 @@ Participant mode:
...
-Intermediaries respect the order of opaque CAPSULE frames: if an intermediary
-receives two opaque CAPSULE frames in a given order, it MUST forward them in
-the same order.
+Capsules MUST be forwarded unmodified by intermediaries, with the exception of
+the DATAGRAM capsule; the DATAGRAM capsule MAY be parsed, added, or removed by
+intermediaries, see {{datagram-capsule}}. Definitions of new Capsule Types MAY
+specify custom intermediary processing.
 
 Endpoints which receive a Capsule with an unknown Capsule Type MUST silently
 drop that Capsule.
```
martinthomson, line 238: This doesn't belong in intermediary processing.  Especially when it might be interpreted as applying to intermediaries.

draft-ietf-masque-h3-datagram.md
```diff
@@ -303,6 +193,10 @@ Capsule Value:
 : The payload of this capsule. Its semantics are determined by the value of the
 Capsule Type field.
 
+Note that use of the Capsule Protocol is not required to use HTTP Datagrams.
+Definitions of new HTTP Methods or of new HTTP Upgrade Tokens can use HTTP
+Datagrams with their own data stream protocol.
+
 
 ## Requirements
```
martinthomson, line 202: > A server MUST NOT send any Transfer-Encoding or Content-Length header fields in a 2xx (Successful) response. If a client receives a Content-Length or Transfer-Encoding header fields in a successful response, it MUST treat that response as malformed.

If this is just about the response you need to say that:

1. the capsule protocol is not in use for a response unless it includes a 2xx status code
2. the capsule protocol MUST NOT be used with messages that contain Content-Length or Transfer-Encoding

You also say:

> A request message does not have content.

There are cases where Content-Length can be zero, but that is limited, not universal.  It's only CONNECT that gets that special handling.  For a new method, the capsule protocol will need to be used for the request body, which means that the request will have content.

> A successful response message does not have content.

Again, this is not going to be the case for anything other than CONNECT.

> Responses are not cacheable.

Generally true, but not necessarily universally true.  I think that this is more of a recommendation rather than a requirement.  It doesn't make sense to cache a response that uses the capsule protocol, but who are we to say that?

draft-ietf-masque-h3-datagram.md
```diff
@@ -335,238 +229,50 @@ Participant mode:
...
-also apply to HTTP Datagrams sent and received using the datagram capsules.
+also apply to HTTP Datagrams sent and received using the DATAGRAM capsule.
 
-The datagram capsules are transparent to intermediaries, meaning that
-intermediaries MAY parse them and send datagram capsules that they did not
-receive. This allows an intermediary to reencode HTTP Datagrams as it forwards
-them: in other words, an intermediary MAY send a datagram capsule to forward an
-HTTP Datagram which was received in a QUIC DATAGRAM frame, and vice versa.
+Intermediaries MAY parse DATAGRAM capsules and MAY send DATAGRAM capsules that
+they did not receive. This allows an intermediary to reencode HTTP Datagrams as
```
martinthomson, line 268: "MAY send DATAGRAM capsules that they did not receive" is problematic.  I don't think that we want intermediaries injecting their own special wisdom into the flow, even if that might be possible, it's not what we want to specifically allow.

What you are trying to say here is in the "in other words" clause: that an intermediary may translate QUIC DATAGRAM frames into HTTP DATAGRAM capsules and vice versa.

draft-ietf-masque-h3-datagram.md
```diff
@@ -335,238 +229,50 @@ Participant mode:
...
+forward an HTTP Datagram which was received in a QUIC DATAGRAM frame, and vice
+versa.
 
-Note that while datagram capsules are sent on a stream, intermediaries can
-reencode HTTP Datagrams into QUIC DATAGRAM frames over the next hop, and those
-could be dropped. Because of this, applications have to always consider HTTP
-Datagrams to be unreliable, even if they were initially sent in a capsule.
+Note that while DATAGRAM capsules are sent on a stream, intermediaries can
+reencode DATAGRAM capsules into QUIC DATAGRAM frames over the next hop, and
+those could be dropped.
```
martinthomson, line 275:
```suggestion
Note that while DATAGRAM capsules that are sent on a stream are reliably delivered in order, intermediaries can
reencode DATAGRAM capsules into QUIC DATAGRAM frames when forwarding messages, which could result in loss or reordering.
```

draft-ietf-masque-h3-datagram.md
```diff
@@ -577,12 +283,22 @@ size advertised on that connection is too low), the intermediary SHOULD drop
...
 
 
+## The RELIABLE_DATAGRAM Capsule
+
+This document defines the RELIABLE_DATAGRAM capsule type (see {{iana-types}} for
+the value of the capsule type). This capsule has the same format and semantics
+as the DATAGRAM capsules, however intermediaries MUST NOT convert between
+RELIABLE_DATAGRAM caspules and other encodings of HTTP Datagrams, or vice versa.
+Endpoints MUST treat receipt of a RELIABLE_DATAGRAM capsule as equivalent to
+receipt of a DATAGRAM capsule with the same contents.
```
martinthomson, line 299: I'm fairly sure that we didn't agree to this bit.

On the contrary, I think that several of us opposed it.  Can we cut it and take it up separately?

draft-ietf-masque-h3-datagram.md
```diff
@@ -623,69 +339,24 @@ recent draft version from the intersection set. This ensures that both
...
-Endpoints indicate their support for datagram contexts by sending the
-Sec-Use-Datagram-Contexts header with a value of ?1. If the header is missing
-or has a value different from ?1, that indicates that its sender does not wish
-to use datagram contexts. Endpoints that wish to use datagram contexts SHALL
-send the Sec-Use-Datagram-Contexts header with a value of ?1 on requests and
-responses that use the capsule protocol.
+Endpoints indicate that the Capsule Protocol is in use on the data stream by
+sending the Sec-Capsule-Protocol header with a value of ?1. Intermediaries MAY
+use this header to allow processing of HTTP Datagrams for unknown HTTP methods
+or unknown HTTP Upgrade Tokens.
```
martinthomson, line 351: What if this appears on a response that doesn't have a 2xx status?  What about 1xx?

draft-ietf-masque-h3-datagram.md
```diff
@@ -623,69 +339,24 @@ recent draft version from the intersection set. This ensures that both
 endpoints negotiate the same draft version.
 
 
-# The Sec-Use-Datagram-Contexts HTTP Header {#context-hdr}
+# The Sec-Capsule-Protocol Header
+
+Definitions of new HTTP Methods or of new HTTP Upgrade Tokens that use the
+Capsule Protocol MAY use the Sec-Capsule-Protocol header to simplify
```
martinthomson, line 345:
```suggestion
Capsule Protocol MAY use the Sec-Capsule-Protocol header field to simplify
```

...and throughout.

draft-ietf-masque-h3-datagram.md
```diff
@@ -623,69 +339,24 @@ recent draft version from the intersection set. This ensures that both
...
-responses that use the capsule protocol.
+Endpoints indicate that the Capsule Protocol is in use on the data stream by
+sending the Sec-Capsule-Protocol header with a value of ?1. Intermediaries MAY
+use this header to allow processing of HTTP Datagrams for unknown HTTP methods
+or unknown HTTP Upgrade Tokens.
 
-"Sec-Use-Datagram-Contexts" is an Item Structured Header {{!RFC8941}}. Its
-value MUST be a Boolean, its ABNF is:
+"Sec-Capsule-Protocol" is an Item Structured Header {{!RFC8941}}. Its value MUST
+be a Boolean, its ABNF is:
```
martinthomson, line 354: This bit where you can't define the field as a Boolean directly is weird.  @mnot, why not?

Also, I think that you need to define what to do with multiple values and parameters.

draft-ietf-masque-h3-datagram.md
```diff
@@ -1113,15 +512,12 @@ STREAM(44): HEADERS            -------->
...
   :path = /hello
   :authority = webtransport.example.org:443
-  Origin = https://www.example.org:443
-
-STREAM(44): DATA                -------->
-  Capsule Type = REGISTER_DATAGRAM
-  Datagram Format Type = WEBTRANSPORT_DATAGRAM
-  Datagram Format Additional Data = ""
+  origin = https://www.example.org:443
+  sec-capsule-protocol = ?1
```
martinthomson, line 516: It would be good to have an example that uses capsules but does not include this header field.

martinthomson (changes requested):
Mostly nits here, this is a pretty good set of changes on the whole and the simplifications are definitely welcome.  The reliable datagram thing is the only thing that I think is concretely problematic.

(I also made comments on surrounding text because I reviewed the full document; feel free to ignore those for this and ask me to open new issues.)

draft-ietf-masque-h3-datagram.md
tfpauly: I'm with Martin that I do not want to see this RELIABLE_DATAGRAM defined, especially not in this PR.

Especially given that capsules can't be modified by intermediaries, etc, I don't see the value in defining this here. If protocols want to define another capsule they use to transmit data over their stream of capsules, great—go ahead and define it there!

draft-ietf-masque-h3-datagram.md
tfpauly: I think this is just the standard suggested text for structured fields 🤷

draft-ietf-masque-h3-datagram.md
tfpauly: Agreed

draft-ietf-masque-h3-datagram.md
bemasc, line 235:
```suggestion
specify optional custom intermediary processing.
```

If you have capsule types that cannot safely be passed verbatim by an intermediary, I don't think you're using The Capsule Protocol.

draft-ietf-masque-h3-datagram.md
```diff
@@ -623,69 +339,24 @@ recent draft version from the intersection set. This ensures that both
 endpoints negotiate the same draft version.
 
 
-# The Sec-Use-Datagram-Contexts HTTP Header {#context-hdr}
+# The Sec-Capsule-Protocol Header
```
bemasc, line 342: Why does this have a Sec- prefix?  Why shouldn't Javascript be allowed to set this header on requests it generates?  It seems safe to me.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Agreed, made the change.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Agreed. I refactored the capsule section to contain more information up front.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Agreed. I refactored the capsule section to contain more information up front.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Agreed, moved the requirement on endpoints to the main capsule section

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Agreed, made the changes

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Clarified that Sec-Capsule-Protocol is only allowed on 2xx

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Yes, this is standard boilerplate.

I added text about multiple values and parameters.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: I removed the header from the WebTransport example.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Agreed, added optional.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: That sounds dangerous, if we allow JavaScript to send this we're adding a burden on every new capsule type to consider the implications of that. I'd rather disallow it to limit risk and avoid potential issues. If someone has a use-case that requires this we could reconsider that choice.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Agreed, rephrased the text

draft-ietf-masque-h3-datagram.md
DavidSchinazi: I'd like us to discuss this in the context of this design team. The reason for that is that it has significant implications on future extensibility. I feel strongly that this spec needs enough of a toehold so that I can write a separate "datagram contexts" extension that has the "zero-latency extensibility" feature. The minimal toehold that I wrote initially was adding an UNEXTENDED_DATAGRAM capsule that has the same semantics as the DATAGRAM capsule but that intermediaries can't touch. Looking back I realized that this ended up being very similar to the RELIABLE_DATAGRAM that Ben asked for so I renamed it. I wrote up the way the extensibility toehold works in the "Creating a Demultiplexing Extension" appendix. I'm definitely open to other toehold that allow future extensibility, let me know if you have ideas.

afrind: I see some generic utility of RELIABLE_DATAGRAM.  It's, well, reliable, and is not bound by MTU restrictions.  The only difference between RELIABLE_DATAGRAM and an unknown Capsule type is that an endpoint has to treat it the same as receiving a DATAGRAM Capsule, instead of dropping it.

I'm missing how having it enables zero-latency extensibility.

draft-ietf-masque-h3-datagram.md
```diff
@@ -303,270 +199,55 @@ Capsule Value:
...
-Types as opaque.
-
-Intermediaries respect the order of opaque CAPSULE frames: if an intermediary
-receives two opaque CAPSULE frames in a given order, it MUST forward them in
-the same order.
+Capsules MUST be forwarded unmodified by intermediaries, with the exception of
+the DATAGRAM capsule. An intermediary that understands the request semantics
+enough to know that capsules are in use MAY parse, add, or remove DATAGRAM
+capsules; see {{datagram-capsule}}. Definitions of new Capsule Types MAY specify
+optional custom intermediary processing.
```
ekr, line 206: This paragraph still concerns me. I'm happy to acknowledge that intermediaries exist, but I dont think we should be specifying their behavior. What happens if we just delete it?

draft-ietf-masque-h3-datagram.md
ekr: I am also skeptical of this. 

I would ask if this question really needs to be resolved by this design team. Can't we discuss it in the WG?

draft-ietf-masque-h3-datagram.md
```diff
@@ -623,68 +314,30 @@ recent draft version from the intersection set. This ensures that both
...
+Definitions of new HTTP Methods or of new HTTP Upgrade Tokens that use the
+Capsule Protocol MAY use the Sec-Capsule-Protocol header field to simplify
+intermediary processing.
 
-"Sec-Use-Datagram-Contexts" is an Item Structured Header {{!RFC8941}}. Its
-value MUST be a Boolean, its ABNF is:
+Endpoints indicate that the Capsule Protocol is in use on the data stream by
+sending the Sec-Capsule-Protocol header field with a value of ?1. Intermediaries
+MAY use this header field to allow processing of HTTP Datagrams for unknown HTTP
+methods or unknown HTTP Upgrade Tokens.
```
ekr, line 326: This is more material specifying intermediaries.

draft-ietf-masque-h3-datagram.md
```diff
@@ -623,68 +314,30 @@ recent draft version from the intersection set. This ensures that both
...
+
+"Sec-Capsule-Protocol" is an Item Structured Header {{!RFC8941}}. Its value MUST
+be a Boolean, its ABNF is:
 
 ~~~
-Sec-Use-Datagram-Contexts = sf-boolean
+Sec-Capsule-Protocol = sf-boolean
```
ekr, line 332: What are the semantics of this boolean? I don't see them defined.

ekr (commented):
I agree that this is simpler, but I still am unpersuaded that we need to specify the behavior of intermediaries as this does.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: If we don't specify intermediary handling, intermediary implementors will implement what they feel like, such as dropping unknown capsules. If we do that we can't use capsules for end to end messages like we do in WebTransport and CONNECT-IP.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: For me the extensibility properties are a fundamental part of the design, so the output of the design team needs to discuss this.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: That is correct.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Good point, I've added text to clarify the semantics.

draft-ietf-masque-h3-datagram.md
ekr: > If we don't specify intermediary handling, intermediary implementors will implement what they feel like, such as dropping unknown capsules. 

And the solution is for service operators not to use such intermediaries.


> If we do that we can't use capsules for end to end messages like we do in WebTransport and CONNECT-IP.

No, you can't do them in situations where you don't control the intermediaries.

This whole discussion of the position vis-a-vis intermediaries was a key point of disagreement on the call, so as long as that's unresolved, this PR isn't done.

draft-ietf-masque-h3-datagram.md
martinthomson: I know that we just shove "Sec-" on everything nowadays, but I think we need to meet a higher bar than "sounds dangerous".  Do you have a concrete example?

Let's say that I did a POST using fetch() with this new field attached.  I do that over h2 and a CDN upgrades to h3 toward the origin, splitting out apparent "datagrams" into frames on that leg.  The server might still accept the request, but discard the datagrams.  That means that the CDN and origin server see a different payload.  Maybe that's something that can be exploited.

But then, what is special about that being dependent on JavaScript?  I don't think that it works as an attack. Fetch will preflight the request and will not add the header unless it is specifically allowed by the server in a preflight, see [CORS-unsafe request-header names](https://fetch.spec.whatwg.org/#cors-unsafe-request-header-names).

draft-ietf-masque-h3-datagram.md
DavidSchinazi: I'm not suggesting that this PR is done, resolving questions like this is why we're having this conversation :)

I see two options here:
1. the current text
2. remove mentions of intermediaries and require service operators terminate every protocol at the intermediary

I think (2) is more complex to deploy, and the cost of (1) is a few lines of spec that implementors of non-intermediaries can ignore, so I'm personally leaning towards (1).

draft-ietf-masque-h3-datagram.md
DavidSchinazi: That's fair. You understand CORS better than I do. After thinking about this some more I have no strong opinion here so I'm happy to go with the consensus of the design team and/or WG on whether we should start with "Sec-" or not.

draft-ietf-masque-h3-datagram.md
```diff
@@ -577,12 +258,22 @@ size advertised on that connection is too low), the intermediary SHOULD drop
 the HTTP Datagram instead of converting it to a DATAGRAM capsule. This
 preserves the end-to-end unreliability characteristic that methods such as
 Datagram Packetization Layer Path MTU Discovery (DPLPMTUD) depend on
-{{?RFC8899}}. An intermediary that converts QUIC DATAGRAM frames to datagram
+{{?RFC8899}}. An intermediary that converts QUIC DATAGRAM frames to DATAGRAM
 capsules allows HTTP Datagrams to be arbitrarily large without suffering any
 loss; this can misrepresent the true path properties, defeating methods such a
```
chris-wood, line 263:
```suggestion
loss; this can misrepresent the true path properties, defeating methods such as
```

draft-ietf-masque-h3-datagram.md
```diff
@@ -303,270 +199,55 @@ Capsule Value:
...
-newly introduced type is transparent. Intermediaries MUST treat unknown Capsule
-Types as opaque.
-
-Intermediaries respect the order of opaque CAPSULE frames: if an intermediary
-receives two opaque CAPSULE frames in a given order, it MUST forward them in
-the same order.
+Capsules MUST be forwarded unmodified by intermediaries, with the exception of
+the DATAGRAM capsule. An intermediary that understands the request semantics
+enough to know that capsules are in use MAY parse, add, or remove DATAGRAM
+capsules; see {{datagram-capsule}}. Definitions of new Capsule Types MAY specify
```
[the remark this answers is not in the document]
DavidSchinazi: Agreed, tightened the text.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: I've removed the Sec- prefix.

draft-ietf-masque-h3-datagram.md
afrind: > this is unordered but reliable

How is it unordered if it can't be converted to a QUIC DATAGRAM?

> That said, I agree with Martin that it doesn't belong in this PR. Let's work on one piece at a time

+1 

There's a lot here we have agreement on, so if RELIABLE_DATAGRAM is controversial let's break it out.  

I'm still missing how it enables or is required for 0-RTT extensions (which are a nice-to-have).  The appendix was vague.  @DavidSchinazi can you write down what you are thinking.

DavidSchinazi: The current draft (version -05) contains the datagram context extensibility system. I'm OK with us removing that from this draft (which is what this PR does), on the condition that I can still implement that as an extension to this draft. In particular, I need that extension to work without costing a round trip (what I've been calling "zero-latency extensibility"). Let me go into more detail as to why RELIABLE_DATAGRAM is important to that.

The datagram context extension works as follows: it is negotiated by adding a header to both request and response, and when both endpoints support it, it adds a context ID varint at the start of the HTTP Datagram Payload field.

Now, in order to achieve zero-latency extensibility, the client needs to be able to send datagrams before receiving the server headers in such a way that the server can correctly parse them (if you add the context ID and the server doesn't support the extension, you don't want the server to parse the context ID as part of the payload).

So here's my idea for how to accomplish that. The extension says the following:
- clients that support the extension are not allowed to send QUIC DATAGRAM frames nor DATAGRAM capsule before they've received the server headers
- the extension defines a new RELIABLE_DATAGRAM_EXT capsule that can be sent before and after receiving server headers
```
Extended Reliable Datagram Capsule {
  Type (i) = RELIABLE_DATAGRAM_EXT,
  Length (i),
  Context ID (i),
  HTTP Datagram Payload (..),
}
```
- the extension does not change the semantics of the RELIABLE_DATAGRAM capsule, that still does not contain a Context ID
```
Reliable Datagram Capsule {
  Type (i) = RELIABLE_DATAGRAM,
  Length (i),
  HTTP Datagram Payload (..),
}
```
- when both endpoints support the extension (after receiving headers) the extension changes the semantics of both the QUIC DATAGRAM frame and DATAGRAM capsule to include a Context ID at the start:
```
Datagram Capsule {
  Type (i) = DATAGRAM,
  Length (i),
  Context ID (i),
  HTTP Datagram Payload (..),
}
HTTP/3 Datagram { // contents of QUIC DATAGRAM frame
  Quarter Stream ID (i),
  Context ID (i),
  HTTP Datagram Payload (..),
}
```
- if the client wishes to use zero-latency extensibility, it can send RELIABLE_DATAGRAM and RELIABLE_DATAGRAM_EXT before receiving the server headers. Both of these will be parsed correctly by servers, whether or not the server supports the extension.

All of this relies on the fact that there is one bit that differentiates the DATAGRAM capsule from the RELIABLE_DATAGRAM capsule (the bit is in the capsule type), so that the extension can give the two capsules different semantics in a way that is backwards compatible.

What I like about this design is that the added complexity to the HTTP Datagram spec is as minimal as possible: it just requires two case statements instead of one in the switch statement that parses capsule types.

I'm OK removing the RELIABLE_DATAGRAM capsule from this PR, as long as we have some other mechanism that I can use to write my extension.

draft-ietf-masque-h3-datagram.md
```diff
@@ -303,270 +199,55 @@ Capsule Value:
...
 it is allowed to send an HTTP Datagram and how to process them from {{format}}
-also apply to HTTP Datagrams sent and received using the datagram capsules.
+also apply to HTTP Datagrams sent and received using the DATAGRAM capsule.
 
-The datagram capsules are transparent to intermediaries, meaning that
-intermediaries MAY parse them and send datagram capsules that they did not
-receive. This allows an intermediary to reencode HTTP Datagrams as it forwards
-them: in other words, an intermediary MAY send a datagram capsule to forward an
-HTTP Datagram which was received in a QUIC DATAGRAM frame, and vice versa.
+Intermediary can reencode HTTP Datagrams as it forwards them: in other words, an
```
MikeBishop, line 243:
```suggestion
An intermediary can reencode HTTP Datagrams as it forwards them. In other words, an
```

draft-ietf-masque-h3-datagram.md
achernya: I find this entire conversation a bit hard to follow, so forgive me if I've missed anything.

There's a claim that RELIABLE_DATAGRAM is controversial, and also a poor name. I haven't seen a good argument to remove it, merely the assertion that people do not want it. There are however a few use cases that have been described in its favor, so I personally would like to see it included (even if it has some other name), as per the discussion we've had at IETF and on list around the different MTU issues, as well as the H3->H2->H3 translation issues that it helps resolve.

A different issue that I'd like to raise (although I can live with this) is that I'm not really a fan of the extension-that-changes-the-semantics-of-DATAGRAM scheme that is being proposed here. I understand that is a natural consequence of the Datagram Capsule -> QUIC DATAGRAM encoding, but I'd personally really like to see the type varint make it into the QUIC DATAGRAM so that we can always unambiguously parse the datagrams, regardless of how we received them (capsule or otherwise). I acknowledge that this may not be the way others want this, and can live with the current (proposed) text.

achernya:
Can we remove the "Creating a Demultiplexing Extension" section? I find it really confusing and hard to follow and it's unclear to me what its purpose is -- is "demultiplexing" the h3->h2->h3 problem, or the "identify different kinds of datagrams (like context ID did before)" problem?

draft-ietf-masque-h3-datagram.md
ekr: > There's a claim that RELIABLE_DATAGRAM is controversial, and also a poor name. I haven't seen a good argument to remove it, merely the assertion that people do not want it. 

This is the wrong standard. Things need consensus to be *included*, not removed.

In any case, a comment on this piece of the document is the wrong place to debate that, so the right thing to do is not have it in this PR and then discuss it separately.

achernya: I disagree with your position that it is reasonable to omit things just because someone provides an unsubstantiated objection. Kindly please provide a technical argument for its removal, or we will not be able to build consensus in either direction.

afrind: Maybe I don't understand what 0-RTT extensibility means.  Does it mean:

1. A client that supports an extension can try to use the extension before receiving confirmation that the server supports it
2. If the server supports the extension it will be processed correctly
3. If the server does not support the extension, the stream will not be torn down, but the extended datagrams could be dropped
---
If that's correct, can you build this without RELIABLE_DATAGRAM by:

1. Extension defines new Capsule type DATAGRAM_EXT
2. Client sends indication that it supports DATAGRAM_EXT in the CONNECT headers
3. Client sends DATAGRAM_EXT Capsules
3a.  Server supports extension, processes DATAGRAM_EXT Capsules
4a. Server confirms extension support
5a. Client receives acknowledgement that the extension is supported, switches to DATAGRAM and DATAGRAM Capsules
3b. Server silently drops DATAGRAM_EXT Capsules
4b. Server sends CONNECT response
5b. Client sees the server does not support the extension, uses standard DATAGRAM and DATAGRAM Capsules

I see that RELIABLE_DATAGRAM makes it possible for a server that supports the extension to assert than any DATAGRAM or DATAGRAM Capsule is only the extended flavor.  But how would an extended server handle a non-extended client in that case?  It seems like the extended server has to have a switch to know if the DATAGRAMs are extended or not, so I don't see any value in RELIABLE_DATAGRAM in this use case.  

Please help me understand if I'm missing something.

ekr: The way this process is supposed to work is that *this* PR is supposed to include the points on which we had consensus, and that means when people object to things which clearly don't, they are removed and deferred for later discussion. If that's not going to happen, then we can just abandon this PR and go back to real time discussion.

afrind: > A different issue that I'd like to raise (although I can live with this) is that I'm not really a fan of the extension-that-changes-the-semantics-of-DATAGRAM scheme that is being proposed here.

I am inclined to agree, but I see the solution as moving the demux higher in the protocol stack (eg: to CONNECT-UDP or CONNECT-IP) rather than down (into QUIC or H3 Datagram).  CONNECT-IP can say all DATAGRAMs on a CONNECT-IP stream have a specific format after the stream ID that includes "Context ID", "Format Type", etc, and this doesn't need to be negotiated at all.  I'd slightly prefer if those protocols used less generic concepts in doing so, but can live with it either way.

achernya: ekr: You appear to have a particularly rigid definition of "the way this is supposed to work", and it does not match mine, and I rather suspect many other people's. I am not particularly interested in having a procedural discussion here, but rather working towards consensus on the document. To that end, I would like to see technical reasoning for everything that goes _in_ or gets taken _out_. I do not consider emotional appeals such as "I don't want this" to be valid.

achernya: afrind: The one downside with your suggestion is it means the context sensitivity still exists, but is a property of the protocol rather than an extension. I think that makes logical sense, but doesn't fully solve the issue I was describing. It's probably the case that we don't necessarily have to solve it, as I mentioned before, I can live with the semantic change here.

DavidSchinazi: @afrind regarding your proposal in https://github.com/ietf-wg-masque/draft-ietf-masque-h3-datagram/pull/124#discussion_r753515209 the issue there is that it doesn't allow the client to send an unextended datagram in its first flight. For CONNECT-UDP that's really useful: the client wants to send its first UDP payload with the first flight and have the server still correctly parse that even if it doesn't support the extension.

@ekr and @achernya in terms of process here: it's clear that there isn't consensus to land this PR with RELIABLE_DATAGRAM, and there isn't consensus to land it without RELIABLE_DATAGRAM. We hadn't really discussed RELIABLE_DATAGRAM at our last in-person design team so I wasn't sure what people's thoughts on it were. Clearly it's something that warrants in-person discussion time. My plan is to keep the topic of RELIABLE_DATAGRAM as an open issue that we will resolve at our next in-person design team meeting. This PR won't get merged until this is resolved.

afrind: > the issue there is that it doesn't allow the client to send an unextended datagram in its first flight.

I assume you mean 'unextended DATAGRAM Capsule'.  In that case, I think I understand, which is that a DATAGRAM Capsule can be converted to a QUIC DATAGRAM by an intermediary, and there's ambiguity at a server that supports the extension after it confirms support: when did the client receive confirmation and start sending extended QUIC DATAGRAMs?

The client can still send DATAGRAM Capsules in this case, and move to QUIC DATAGRAM after it knows whether the server supports the extension or not.  Is there a real use case for sending both extended and non-extended DATAGRAMs in 0-RTT, where we're willing to sacrifice HoL blocking resilience and possible flow control constraints for the extended ones, but not willing to do so for the non-extended ones?  Or maybe the use-cases are small but we think the cost of RELIABLE_DATAGRAM is low and maybe has other utility?

This brings me back here:

> I'm not really a fan of the extension-that-changes-the-semantics-of-DATAGRAM scheme that is being proposed here.

I think we've shown that it's pretty hard to negotiate changes to QUIC DATAGRAM formats on the fly.  Let's not do that.  The cost of easy extensibility is an additional byte (or two) of header information inside the datagram payload of protocols that need it.  If CONNECT-UDP needs to have a such a header I can live with that.

DavidSchinazi: QUIC DATAGRAM frames and DATAGRAM capsules need to have the exact same semantics, because intermediaries convert blindly between them so you can't know how they were originally sent. That's why my proposal forbids sending either when you support the extension and haven't yet received the peer's headers.

My personal opinion is that
- the cost of having RELIABLE_DATAGRAM for an implementation that doesn't care is very low
- RELIABLE_DATAGRAM has value in and of itself
- moving the complexity of contexts to their own extension allows defining different ways to do this down the road, whereas if we say that CONNECT-UDP starts with a varint we've committed to this context design forever

Let's discuss this in person on 11/29

martinthomson: Congratulations, you have just collectively managed to do what we've learned through hard experience is a bad idea: a substantive technical discussion in a comment attached to a line of code on a pull request.

All I was asking for was for this to be taken to an issue.  Can we do that?

draft-ietf-masque-h3-datagram.md
```diff
@@ -623,68 +314,30 @@ recent draft version from the intersection set. This ensures that both
...
+intermediary processing.
 
-"Sec-Use-Datagram-Contexts" is an Item Structured Header {{!RFC8941}}. Its
-value MUST be a Boolean, its ABNF is:
+"Capsule-Protocol" is an Item Structured Header {{!RFC8941}}. Its value MUST be
+a Boolean, its ABNF is:
 
 ~~~
-Sec-Use-Datagram-Contexts = sf-boolean
+Capsule-Protocol = sf-boolean
```
vasilvv, line 327: I am inclined to make this `Sec-`, because I don't see a scenario in which JavaScript would need to send it, and in general I believe that new request headers should be defined as `Sec-` unless they're both safe and useful for JavaScript applications.

vasilvv (commented):
LGTM minus figuring out what we want to do with RELIABLE_DATAGRAMS.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Sure! I've filed #125 to discuss this.

draft-ietf-masque-h3-datagram.md
```diff
@@ -37,9 +37,9 @@ with a mechanism to send unreliable data while leveraging the security and
 congestion-control properties of QUIC. However, QUIC DATAGRAM frames do not
 provide a means to demultiplex application contexts. This document describes how
 to use QUIC DATAGRAM frames when the application protocol running over QUIC is
-HTTP/3. It associates datagrams with client-initiated bidirectional streams
-and defines an optional additional demultiplexing layer. Additionally, this
-document defines how to convey datagrams over prior versions of HTTP.
+HTTP/3. It associates datagrams with client-initiated bidirectional streams.
+Additionally, this document defines how to convey datagrams over prior versions
```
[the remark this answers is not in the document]
DavidSchinazi: Agreed, done.

draft-ietf-masque-h3-datagram.md
```diff
@@ -91,85 +80,13 @@ when, and only when, they appear in all capitals, as shown here.
...
-allocation carries separate namespaces to avoid requiring synchronization.
-Additionally, note that the context ID namespace is tied to a given HTTP
-request: it is possible for the same numeral context ID to be used
-simultaneously in distinct requests.
+coexist on a given QUIC connection. To allow this, the QUIC DATAGRAM frame
+payload starts with an encoded stream identifier that associates the datagram
+with a request stream.
+
+When running over HTTP/2, demultiplexing is provided by the HTTP/2 framing
+layer. When running over HTTP/1, requests are strictly serialized in the
```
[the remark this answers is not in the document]
DavidSchinazi: Agreed, done.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: It sounds like we have folks who prefer with `Sec-` and folks who prefer without. Let's discuss in person.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: I've removed RELIABLE_DATAGRAM from this PR. We still have #111 to discuss RELIABLE_DATAGRAM and #125 to discuss zero-latency extensibility, so I'm now marking this comment thread as resolved.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: In today's design team meeting, we reached rough consensus to not use the `Sec-` prefix. We still have [an issue](https://github.com/ietf-wg-masque/draft-ietf-masque-connect-udp/issues/68) to track adding text to Security Considerations, so I'm marking this comment thread as resolved.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: In today's design team meeting, we reached rough consensus to keep the current approach to intermediaries in this PR. We still have [an issue](https://github.com/ietf-wg-masque/draft-ietf-masque-connect-udp/issues/69) to track some changes that we still need to make, so I'm marking this comment thread as resolved.

tfpauly (approved):

achernya (approved):

draft-ietf-masque-h3-datagram.md
```diff
@@ -90,86 +79,16 @@ when, and only when, they appear in all capitals, as shown here.
 
 # Multiplexing
 
+All HTTP Datagrams are associated with an HTTP request stream.
```
LPardue, line 82: request stream, or simply request?

```suggestion
All HTTP Datagrams are associated with an HTTP request.
```

draft-ietf-masque-h3-datagram.md
```diff
@@ -90,86 +79,16 @@ when, and only when, they appear in all capitals, as shown here.
 
 # Multiplexing
```
LPardue, line 80: Strictly editorial: a better way to structure this section would be to describe what we defin multiplexing is. Then the follow on paragraphs state how each version can (or cannot) satisfy that. That would mean simplifying paragraph 2 of this section by extracting the "need the ability" piece so that the text focuses on how not why.

Such an editorial change could be tracked as follow on work, since this PR is focuesd on design changes.

draft-ietf-masque-h3-datagram.md
```diff
@@ -249,23 +141,29 @@ corresponding stream.
...
-communicate request-related information end-to-end, even in the presence of
-HTTP intermediaries.
-
-
-## Capsule Protocol {#capsule-protocol}
+sequence of type-length-value tuples that new HTTP Methods or new HTTP Upgrade
+Tokens can choose to use. It allows endpoints to reliably communicate
+request-related information end-to-end on HTTP request streams, even in the
+presence of HTTP intermediaries. The Capsule Protocol can be used to exchange
+HTTP Datagrams over versions of HTTP that do not support the QUIC DATAGRAM frame.
```
LPardue, line 148: ambiguity alert: do we striclty mean HTTP versions? Presumably HTTP/3 connections that do not support QUIC DATAGRAM (or chose not to use it even if supported) are also in scope.

Consider moving the last sentence here to para 3, so that mention of Capsule Protocol plus HTTP Datagrams is grouped.

draft-ietf-masque-h3-datagram.md
```diff
@@ -249,23 +141,29 @@ corresponding stream.
 # Capsules {#capsule}
```
LPardue, line 141: Editorial: Not strictly related to this PR but consider renaming this section to `Capsule Protocol` and the subsection to something else (possibly `Using the Capsule Protocol")

draft-ietf-masque-h3-datagram.md
```diff
@@ -249,23 +141,29 @@ corresponding stream.
...
 
 This specification defines the "data stream" of an HTTP request as the
 bidirectional stream of bytes that follow the headers in both directions. In
 HTTP/1.x, the data stream consists of all bytes on the connection that follow
 the blank line that concludes either the request header section, or the 2xx
-(Successful) response header section. In HTTP/2 and HTTP/3, the data stream of
-a given HTTP request consists of all bytes sent in DATA frames with the
+(Successful) response header section. In HTTP/2 and HTTP/3, the data stream of a
+given HTTP request consists of all bytes sent in DATA frames with the
 corresponding stream ID. The concept of a data stream is particularly relevant
```
LPardue, line 156: Editorial: you might be able to strengthen the definition of data stream a bit by extracting the semantic from version concerns. For example, something like

"This specification defines the "data stream" of an HTTP request as the
bidirectional stream of bytes that follow the header section in both directions. 
The concept of a data stream is particularly relevant for methods such as 
CONNECT. In the client-to-server direction, the data stream always exists.
In the server-to-client direction, the data stream only exists if when there
is a 2xx (Successful) response. In HTTP/1.x, the data stream consists of all
bytes on the connection that follow the blank line concluding header section.
In HTTP/2 and HTTP/3, the data stream of a given HTTP request consists
of all bytes sent in DATA frames with the corresponding stream ID. "

draft-ietf-masque-h3-datagram.md
```diff
@@ -303,270 +201,55 @@ Capsule Value:
...
-document are opaque, with the exception of the datagram capsules, see
-{{datagram-capsule}}. Definitions of new Capsule Types MAY specify that the
-newly introduced type is transparent. Intermediaries MUST treat unknown Capsule
-Types as opaque.
-
-Intermediaries respect the order of opaque CAPSULE frames: if an intermediary
-receives two opaque CAPSULE frames in a given order, it MUST forward them in
-the same order.
+Capsules MUST be forwarded unmodified by intermediaries, with the exception of
+the DATAGRAM capsule. An intermediary that understands the request semantics
```
LPardue, line 205: editorial: move `;see {{datagram-capsule}}` to the first ref here

draft-ietf-masque-h3-datagram.md
```diff
@@ -303,270 +201,55 @@ Capsule Value:
...
-
-If datagrams contexts are not in use, the client is responsible for choosing
-the datagram format and informing the server via a REGISTER_DATAGRAM capsule.
-Servers MUST NOT send the REGISTER_DATAGRAM capsule. If a client receives a
-REGISTER_DATAGRAM capsule, the client MUST abruptly terminate the corresponding
-stream with a stream error of type H3_GENERAL_PROTOCOL_ERROR.
+By virtue of the definition of the data stream, the Capsule Protocol is not in
+use on responses unless the response includes a 2xx (Successful) status code.
+The Capsule Protocol MUST NOT be used with messages that contain Content-Length
+or Transfer-Encoding header fields.
```
LPardue, line 216: Editorial: Consider moving this paragraph closer to the definition of data stream, so that you can lose the first sentence.

draft-ietf-masque-h3-datagram.md
```diff
@@ -303,270 +201,55 @@ Capsule Value:
...
-
-This document defines the DATAGRAM and DATAGRAM_WITH_CONTEXT capsules types,
-known collectively as the datagram capsules (see {{iana-types}} for the value
-of the capsule types). These capsules allow an endpoint to send a datagram
-frame over an HTTP stream. This is particularly useful when using a version of
-HTTP that does not support QUIC DATAGRAM frames.
+This document defines the DATAGRAM capsule type (see {{iana-types}} for the
+value of the capsule type). This capsule allows an endpoint to send an HTTP
+Datagram on a stream using the Capsule Protocol. This is particularly useful
+when using a version of HTTP that does not support QUIC DATAGRAM frames.
```
LPardue, line 224: see earlier comment about whether this is strictly "version of HTTP" or not.

draft-ietf-masque-h3-datagram.md
```diff
@@ -623,68 +306,30 @@ recent draft version from the intersection set. This ensures that both
 endpoints negotiate the same draft version.
 
 
-# The Sec-Use-Datagram-Contexts HTTP Header {#context-hdr}
+# The Capsule-Protocol Header
```
LPardue, line 309: Editorial: consider moving this whole section to be part of the definition of Capsule protocol definition section

LPardue (approved):
Design-wise: LGTM

I made a bunch of editorial suggestions. These are non-blocking and we can consider them in follow-on work if nobody wants to change this PR up too much.

MikeBishop (approved):
Good direction; some open issues to resolve, but I don't think they block the main direction of the work.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: This sounds good, but I agree - let's tackle this as editors once the outcome of the design team is merged.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Let's tackle that post design team

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Fixed

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Let's tackle that in a follow-up since the design team isn't touching the concept of the data stream

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Done

draft-ietf-masque-h3-datagram.md
DavidSchinazi: The purpose of this is to put requirements on the capsule protocol, so it wouldn't look right if moved to before the capsule protocol isn't introduced.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Fixed

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Done

draft-ietf-masque-h3-datagram.md
```diff
@@ -623,68 +306,30 @@ recent draft version from the intersection set. This ensures that both
...
-in use, and they MAY do so in a way which is opaque to intermediaries.
+Endpoints indicate that the Capsule Protocol is in use on the data stream by
+sending the Capsule-Protocol header field with a value of ?1. A Capsule-Protocol
+header field with a value of ?0 has the same semantics as when the header is not
+present. Intermediaries MAY use this header field to allow processing of HTTP
+Datagrams for unknown HTTP methods or unknown HTTP Upgrade Tokens.
+
+The Capsule-Protocol header field MUST NOT be sent multiple times on a message.
+The Capsule-Protocol header field MUST NOT be used on HTTP responses with a
+status code different from 2xx (Successful). This specification does not define
```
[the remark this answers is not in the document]
DavidSchinazi: Let's take that to issues (I filed #130 and #131), I think we can handle this outside the design team

PR #132, opened 2021-12-02
Clarify meaning of understanding capsule request semantics

### LINKED ISSUE #127: Conditions on datagram forwarding
> An intermediary that understands the request semantics enough to know that capsules are in use MAY convert between QUIC DATAGRAM frames and DATAGRAM capsules

I think that this is right, but we should be more precise about what this means.

That is, if the method is CONNECT and either:

1. the intermediary understands the :protocol selection, or
2. the request includes Capsule-Protocol

Then the intermediary can forward DATAGRAM capsules as QUIC DATAGRAM frames or forward QUIC DATAGRAM frames as DATAGRAM capsules.

Those conditions could be spelled out more clearly.

DavidSchinazi:
Fixes #127.

cc @martinthomson

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -204,9 +204,10 @@ Capsule Type field.
 
 Capsules MUST be forwarded unmodified by intermediaries, with the exception of
 the DATAGRAM capsule; see {{datagram-capsule}}. An intermediary that understands
-the request semantics enough to know that capsules are in use MAY convert
-between QUIC DATAGRAM frames and DATAGRAM capsules. Definitions of new Capsule
-Types MAY specify optional custom intermediary processing.
+the request semantics enough to know that capsules are in use (either by
+recognizing the HTTP Upgrade token, or the Capsule-Protocol header from {{hdr}})
+MAY convert between QUIC DATAGRAM frames and DATAGRAM capsules. Definitions of
+new Capsule Types MAY specify optional custom intermediary processing.
 
 Endpoints which receive a Capsule with an unknown Capsule Type MUST silently
 drop that Capsule.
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-h3-datagram.md
```diff
@@ -204,9 +204,10 @@ Capsule Type field.
 
 Capsules MUST be forwarded unmodified by intermediaries, with the exception of
 the DATAGRAM capsule; see {{datagram-capsule}}. An intermediary that understands
-the request semantics enough to know that capsules are in use MAY convert
-between QUIC DATAGRAM frames and DATAGRAM capsules. Definitions of new Capsule
-Types MAY specify optional custom intermediary processing.
+the request semantics enough to know that capsules are in use (either by
+recognizing the HTTP Upgrade token, or the Capsule-Protocol header from {{hdr}})
```
martinthomson, line 208:
```suggestion
recognizing the HTTP Upgrade token or the Capsule-Protocol header from {{hdr}})
```

draft-ietf-masque-h3-datagram.md
```diff
@@ -204,9 +204,10 @@ Capsule Type field.
...
+MAY convert between QUIC DATAGRAM frames and DATAGRAM capsules. Definitions of
```
martinthomson, line 209: I might prefer two discrete sentences:

> An intermediary can identify the use of the capsule protocol either through the presence of the Capsule-Protocol header field ({{hdr}}) or by understanding the chosen HTTP Upgrade token. An intermediary that identifies the use of the capsule protocol MAY convert between DATAGRAM capsules and QUIC DATAGRAM frames when forwarding.

DavidSchinazi: Sounds good to me, done

PR #135, opened 2022-01-21
HTTP Datagrams: 2021 Design Team Consensus

### LINKED ISSUE #84: Remove context identifiers
The capsules that signal use of different datagram contexts (or the absence thereof) assume that there is a generic need for this form of identification.  That has not been established.  WebTransport certainly doesn't need it.  

As this is an *end-to-end* signal anyway, it is also a *protocol-specific* signal.  All uses of this framework are perfectly capable of doing their own signaling and negotiation.  A one-size-fits-all mechanism is more likely a one-size-fits-few mechanism, but it still comes with costs that every user has to bear (see ietf-wg-webtrans/draft-ietf-webtrans-http3#54 for instance).

For instance, if CONNECT-UDP wanted to signal ECN and DSCP, that might be more efficiently achieved with a single byte (2 + 6) rather than a varint.  If CONNECT-IP wanted different compression contexts, that might be suited to a varint, but why would this draft be in a position to decide that?

As there is no value in an intermediary looking at these context identifiers - I would argue negative utility in the RFC 8558 sense - these should be removed from the protocol.

The only potential value I see in having some sort of signaling here is to explicitly signal the use of DATAGRAM frames associated with a stream.  However, as we have agreed to send DATAGRAM frames without negotiation, that too seems like it won't be needed.

### LINKED ISSUE #107: Capsule lengths
The generic definition of Capsule includes the `Capsule Length` field that is the length of all remaining fields, which depends on the type. We state that the length can be zero.

However, many of the capsule types we define cannot be zero-length:

* REGISTER_DATAGRAM - at least 1 byte
* REGISTER_DATAGRAM_CONTEXT - at least 2 bytes
* Datagram Close - at least 2 bytes

We could be more prescriptive of these capsules. Such as explicitly marking the minimal length in the figures, or noting it in the prose. We might also recommend that receivers of these capsules reject capsules that are clearly too short

### LINKED ISSUE #117: Acknowledge that Context ID is optional
> Endpoints parse both the Quarter Stream ID field and the Context ID
   field in order to associate the QUIC DATAGRAM frame with a stream and
   context within that stream.  If an endpoint receives a QUIC DATAGRAM
   frame whose payload is too short to allow parsing the Quarter Stream
   ID field, the endpoint MUST treat it as an HTTP/3 connection error of
   type H3_GENERAL_PROTOCOL_ERROR.  If an endpoint receives a QUIC
   DATAGRAM frame whose payload is long enough to allow parsing the
   Quarter Stream ID field but too short to allow parsing the Context ID
   field, the endpoint MUST abruptly terminate the corresponding stream
   with a stream error of type H3_GENERAL_PROTOCOL_ERROR.

This tests talks about paring of Context ID on an endpoint, but Context ID is optional. The text should acknowledge that better.

### LINKED ISSUE #118: Improve text about multiple CLOSE_DATAGRAM_CONTEXTs
> Endpoints MUST NOT close a Context ID that was not previously
   registered.  Endpoints MUST NOT close a Context ID that has already
   been closed.  If an endpoint receives a CLOSE_DATAGRAM_CONTEXT
   capsule that violates one or more of these requirements, the endpoint
   MUST abruptly terminate the corresponding stream with a stream error
   of type H3_GENERAL_PROTOCOL_ERROR.

I think this text should acknowledge that both sides may close a context at the same time and that a CLOSE_DATAGRAM_CONTEXT may be received for a already closed context. This is just a nit, but I think worth mentioning.

### LINKED ISSUE #125: Zero-Latency Extensibility
We've been discussing this topic on the [design team PR](https://github.com/ietf-wg-masque/draft-ietf-masque-h3-datagram/pull/124#discussion_r751879354) and on the [RELIABLE_DATAGRAM issue](https://github.com/ietf-wg-masque/draft-ietf-masque-h3-datagram/issues/111), but @martinthomson rightly points out that this deserves its own issue.

# Motivation

There is interest in deploying extensions to protocols that use HTTP Datagrams. Some of these extensions require the ability to send multiple types of HTTP datagrams associated with a single request stream. For example,
* an extension to CONNECT-UDP that allows measuring the HTTP Datagram MTU
* an extension to CONNECT-IP that allows compressing IP headers

This means that when these extensions are in use, HTTP Datagrams carry an identifier that allows the receiver to know what kind of HTTP Datagram it's receiving (e.g. "is this a UDP payload or an MTU probe?" or "is this a full IP packet or a compressed one?").

One potential solution is to have CONNECT-UDP and CONNECT-IP require their HTTP Datagrams to always start with an identifier. This has the unfortunate downside of wasting a byte for requests that do not use extensions, and also precludes WebTransport from using extensions in the future.

Alternatively,  we could define an optional extension to HTTP Datagrams which has the advantage of not wasting bytes when not in use, and allows any protocol relying on HTTP Datagrams to use it if chooses to at any point in the future. This extension would be negotiated via an HTTP header.

# Requirement: Zero-Latency Extensibility

Let's assume we are going with the model of an optional extension to HTTP Datagrams. Since this is optional, a client will not know whether it is supported by the server until it receives the response headers. "Zero-Latency Extensibility" is the ability for a client to send HTTP Datagrams (both the "regular" datagram that's the default one from the request, and the "extension" datagrams). Note that this doesn't apply to the server because the server doesn't send HTTP Datagrams until after it's parsed the request headers. Let's call the flight of frames and/or packets that go alongside the HTTP request headers the "first flight". The requirement is for the client to be able to send both "regular" and "extension" HTTP datagrams with its first flight, and to have the server parse them as follows: if the server supports the extension, the server can correctly parse both "regular" and "extension" HTTP datagrams from the first flight; if the server doesn't support the extension, the server can correctly parse the "regular" datagrams and drop the "extension" ones.

# Requirement: Unmodified Intermediaries

Intermediaries shouldn't have to know or care about this extension. All that intermediaries need to do is blindly convert between QUIC DATAGRAM frames and DATAGRAM capsules regardless of any extension being used. This puts one constraint on the solution: QUIC DATAGRAM frames and DATAGRAM capsules need to have the same semantics because they can get converted between one another by intermediaries.

# Solution

Solving this problem once the headers is exchanged is pretty straightforward: if both endpoints support the extension, then datagrams start with an identifier. Otherwise, they don't contain the identifier. Getting this to work before the response headers are received is slightly more tricky: it requires having a way to distinguish between datagrams sent before the headers were received and ones sent after. The simplest way to distinguish them is to define two capsules and use the capsule type to tell us which is which.

The datagram context extension works as follows: it is negotiated by adding a header to both request and response, and when both endpoints support it, it adds a context ID varint at the start of the HTTP Datagram Payload field.

Now, in order to achieve zero-latency extensibility, the client needs to be able to send datagrams before receiving the server headers in such a way that the server can correctly parse them (if you add the context ID and the server doesn't support the extension, you don't want the server to parse the context ID as part of the payload).

So here's my idea for how to accomplish that. The extension says the following:
- clients that support the extension are not allowed to send QUIC DATAGRAM frames nor DATAGRAM capsule before they've received the server headers
- the extension defines a new RELIABLE_DATAGRAM_EXT capsule that can be sent before and after receiving server headers
```
Extended Reliable Datagram Capsule {
  Type (i) = RELIABLE_DATAGRAM_EXT,
  Length (i),
  Context ID (i),
  HTTP Datagram Payload (..),
}
```
- the extension does not change the semantics of the RELIABLE_DATAGRAM capsule, that still does not contain a Context ID
```
Reliable Datagram Capsule {
  Type (i) = RELIABLE_DATAGRAM,
  Length (i),
  HTTP Datagram Payload (..),
}
```
- when both endpoints support the extension (after receiving headers) the extension changes the semantics of both the QUIC DATAGRAM frame and DATAGRAM capsule to include a Context ID at the start:
```
Datagram Capsule {
  Type (i) = DATAGRAM,
  Length (i),
  Context ID (i),
  HTTP Datagram Payload (..),
}
HTTP/3 Datagram { // contents of QUIC DATAGRAM frame
  Quarter Stream ID (i),
  Context ID (i),
  HTTP Datagram Payload (..),
}
```
- if the client wishes to use zero-latency extensibility, it can send RELIABLE_DATAGRAM and RELIABLE_DATAGRAM_EXT before receiving the server headers. Both of these will be parsed correctly by servers, whether or not the server supports the extension.

All of this relies on the fact that there is one bit that differentiates the DATAGRAM capsule from the RELIABLE_DATAGRAM capsule (the bit is in the capsule type), so that the extension can give the two capsules different semantics in a way that is backwards compatible.

I'm realizing that calling this RELIABLE_DATAGRAM tied zero-latency extensibility to #111. This would work just as well if we renamed RELIABLE_DATAGRAM to EARLY_DATAGRAM or UNEXTENDED_DATAGRAM.

### LINKED ISSUE #126: Handling of unrecognized capsule types
With #124 and https://github.com/ietf-wg-masque/draft-ietf-masque-connect-udp/pull/70, there is no longer a CLOSE_CONTEXT_ID Capsule, with its associated error codes.  H3 Datagram still says that unknown capsule types must be ignored.  This creates a dilemma: if I send a capsule like REGISTER_FOO_CONTEXT, acknowledged by a REGISTER_FOO_ACK response, how do I know if it worked?  If the REGISTER_FOO_CONTEXT capsule is ignored, I will have to wait indefinitely for the ACK, or set a timeout (yuck).

There are a few different ways to solve this, but I think the cleanest is the REGISTER_DATAGRAM_CONTEXT capsule type from the (published) H3 Datagram draft.  CONNECT-IP and CONNECT-UDP can make that capsule type mandatory to implement, so we don't need to worry about whether the other side supports it.

This also highlights why we need a "Datagram Contexts" draft supporting CONNECT-UDP and CONNECT-IP.  We need the "Datagram Format Types" concept, the CLOSE_DATAGRAM_CONTEXT capsule type, and all the other bits.  We could paste all of this into CONNECT-UDP, but I don't think we want to cross-reference it out of CONNECT-UDP every time we need to define a new extension or format.

### LINKED ISSUE #127: Conditions on datagram forwarding

DavidSchinazi:
This PR represents the consensus of the 2021 MASQUE Design Team.

[Rendered view of this PR](https://ietf-wg-masque.github.io/draft-ietf-masque-h3-datagram/2021_design_team_consensus/draft-ietf-masque-h3-datagram.html).
[Rendered diff with the editor's copy](https://www.ietf.org/rfcdiff?url1=https://ietf-wg-masque.github.io/draft-ietf-masque-h3-datagram/draft-ietf-masque-h3-datagram.txt&url2=https://ietf-wg-masque.github.io/draft-ietf-masque-h3-datagram/2021_design_team_consensus/draft-ietf-masque-h3-datagram.txt).

cc @afrind @LPardue @MikeBishop @martinthomson @vasilvv @ekr @achernya @bemasc @tfpauly @ihlar

Fixes #84
Fixes #107
Fixes #117 
Fixes #118
Fixes #125
Fixes #126
Fixes #127

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -206,13 +206,15 @@ Capsule Value:
 : The payload of this capsule. Its semantics are determined by the value of the
 Capsule Type field.
 
-Capsules MUST be forwarded unmodified by intermediaries, with the exception of
-the DATAGRAM capsule; see {{datagram-capsule}}. An intermediary can identify the
-use of the capsule protocol either through the presence of the Capsule-Protocol
-header field ({{hdr}}) or by understanding the chosen HTTP Upgrade token. An
-intermediary that identifies the use of the capsule protocol MAY convert between
-DATAGRAM capsules and QUIC DATAGRAM frames when forwarding. Definitions of new
-Capsule Types MAY specify optional custom intermediary processing.
+Because new protocols or extensions may involve defining new capsule types,
+intermediaries that wish to allow for future extensibility SHOULD forward
+capsules unmodified. One exception to this rule is the DATAGRAM capsule; see
+{{datagram-capsule}}. An intermediary can identify the use of the capsule
+protocol either through the presence of the Capsule-Protocol header field
+({{hdr}}) or by understanding the chosen HTTP Upgrade token. An intermediary
+that identifies the use of the capsule protocol MAY convert between DATAGRAM
+capsules and QUIC DATAGRAM frames when forwarding. Definitions of new Capsule
+Types MAY specify optional custom intermediary processing.
 
 Endpoints which receive a Capsule with an unknown Capsule Type MUST silently
 drop that Capsule.
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -229,7 +229,7 @@ intermediary processing.
 a Boolean. Its ABNF is:
 
 ~~~
-Capsule-Protocol = sf-boolean
+Capsule-Protocol = sf-boolean parameters
 ~~~
 
 Endpoints indicate that the Capsule Protocol is in use on the data stream by
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -158,9 +158,13 @@ corresponding stream ID. The concept of a data stream is particularly relevant
 for methods such as CONNECT where there is no HTTP message content after the
 headers.
 
-Note that use of the Capsule Protocol is not required to use HTTP Datagrams.
-Definitions of new HTTP Methods or of new HTTP Upgrade Tokens can use HTTP
-Datagrams with their own data stream protocol.
+Note that use of the Capsule Protocol is not required to use HTTP Datagrams. If
+a new HTTP Method or Upgrade Token is only defined over transports that support
+QUIC DATAGRAM frames, they might not need a stream encoding. Additionally,
+definitions of new HTTP Methods or of new HTTP Upgrade Tokens can use HTTP
+Datagrams with their own data stream protocol. However, new HTTP Method or
+Upgrade Tokens SHOULD use the Capsule Protocol unless they have a good reason
+not to.
 
 
 ## Capsule Protocol {#capsule-protocol}
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -473,7 +473,7 @@ Client                                             Server
 STREAM(44): HEADERS            -------->
   :method = CONNECT
   :scheme = https
-  :method = webtransport
+  :protocol = webtransport
   :path = /hello
   :authority = webtransport.example.org:443
   origin = https://www.example.org:443


LPardue (approved):
LGTM, thank you!

draft-ietf-masque-h3-datagram.md
```diff
@@ -303,270 +202,84 @@ Capsule Value:
...
+header field with a value of ?0 has the same semantics as when the header is not
+present. Intermediaries MAY use this header field to allow processing of HTTP
+Datagrams for unknown HTTP methods or unknown HTTP Upgrade Tokens.
 
-NO_ERROR:
+The Capsule-Protocol header field MUST NOT be sent multiple times on a message.
+The Capsule-Protocol header field MUST NOT be used on HTTP responses with a
+status code different from 2xx (Successful). This specification does not define
+any parameters for the Capsule-Protocol header field value, but future documents
+MAY define parameters. Receivers MUST ignore unknown parameters.
```
vasilvv, line 245: I am confused by this text.  sf-boolean is literally defined to be either `?0` or `?1`, there are no parameters involved.

draft-ietf-masque-h3-datagram.md
```diff
@@ -1113,12 +476,7 @@ STREAM(44): HEADERS            -------->
   :method = webtransport
```
vasilvv, line 476: I think you mean `:protocol = webtransport`, not `:method = webtransport`.  Also, `capsule-protocol = ?1`?

vasilvv (approved):
LGTM.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Fixed `:protocol`. Regarding `capsule-protocol`, we had decided that one of the examples could go without it to indicate that use of the header is optional

draft-ietf-masque-h3-datagram.md
DavidSchinazi: sf-boolean is a subclass of sf-item which allows parameters
https://datatracker.ietf.org/doc/html/rfc8941#section-3.1.2

tfpauly (approved):

draft-ietf-masque-h3-datagram.md
```diff
@@ -249,23 +141,30 @@ corresponding stream.
...
-a given HTTP request consists of all bytes sent in DATA frames with the
+(Successful) response header section. In HTTP/2 and HTTP/3, the data stream of a
+given HTTP request consists of all bytes sent in DATA frames with the
 corresponding stream ID. The concept of a data stream is particularly relevant
 for methods such as CONNECT where there is no HTTP message content after the
 headers.
 
+Note that use of the Capsule Protocol is not required to use HTTP Datagrams.
+Definitions of new HTTP Methods or of new HTTP Upgrade Tokens can use HTTP
+Datagrams with their own data stream protocol.
```
ekr, line 163: This text feels like it could use some work. It seems like the situation is as follows:

1. You an use HTTP datagrams without capsules or any other encapsulation protocol if the transport supports QUIC datagram.
2. If the protocol doesn't support QUIC datagram you need some kind of encapsulation protocol. You are free to use capsules or define your own thing in this case.

Is this right?

draft-ietf-masque-h3-datagram.md
```diff
@@ -303,270 +202,84 @@ Capsule Value:
...
-Intermediaries respect the order of opaque CAPSULE frames: if an intermediary
-receives two opaque CAPSULE frames in a given order, it MUST forward them in
-the same order.
+Capsules MUST be forwarded unmodified by intermediaries, with the exception of
+the DATAGRAM capsule; see {{datagram-capsule}}. An intermediary can identify the
+use of the capsule protocol either through the presence of the Capsule-Protocol
+header field ({{hdr}}) or by understanding the chosen HTTP Upgrade token. An
+intermediary that identifies the use of the capsule protocol MAY convert between
+DATAGRAM capsules and QUIC DATAGRAM frames when forwarding. Definitions of new
+Capsule Types MAY specify optional custom intermediary processing.
```
ekr, line 211: I don't recall us having consensus on this text around intermediaries. Can you please point me to the minutes or email thread where we agreed on this?

ekr (changes requested):

draft-ietf-masque-h3-datagram.md
DavidSchinazi: We reached rough consensus on this at the second design team meeting on 2021-11-29: https://github.com/ietf-wg-masque/draft-ietf-masque-h3-datagram/pull/124#discussion_r758842704

draft-ietf-masque-h3-datagram.md
DavidSchinazi: I think that's right, though I'd say that the spec nudges folks towards using capsules, while acknowledging that it's possible to bring your own encapsulation

draft-ietf-masque-h3-datagram.md
ekr: This is rather less than minutes. I'll send mail.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: I wouldn't call the capsule protocol unnecessary, as it has value outside of datagrams: for example WebTransport uses it to send close details, and CONNECT-IP uses it to exchange configuration information.

ekr: OK, well, feel free to revise my proposed text. However, the existing text needs revision.

DavidSchinazi: Done

achernya (approved):
I'm happy with this as-is, but would also support the editorial changes being discussed on the email thread around how intermediaries handle datagrams.

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Discussed over email, and resolved by making the intermediary requirement a SHOULD with explanations

PR #138, opened 2022-02-25
Revisit sticking out text now that document has evolved

### LINKED ISSUE #109: Commentary on "sticking out" is stale and possibly mislocated
The doc has evolved from just HTTP/3 Datagram, to define Capsule protocol and HTTP datagrams over capsule.

In Security Considerations, we say

> Since this feature requires sending an HTTP/3 Settings parameter, it "sticks out".

It's a little ambiguous what "this feature" on face value. Arguably, the recommendation to always send the setting could be done in the section that defines the setting. Some editorial work would address these concerns.

DavidSchinazi:
Fixes #109

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -355,11 +355,12 @@ allow endpoints to communicate their prioritization preferences.
 
 # Security Considerations {#security}
 
-Since this feature requires sending an HTTP/3 Settings parameter, it "sticks
-out". In other words, probing clients can learn whether a server supports this
-feature. Implementations that support this feature SHOULD always send this
-Settings parameter to avoid leaking the fact that there are applications using
-HTTP/3 datagrams enabled on this endpoint.
+Since transmitting HTTP Datagrams using QUIC DATAGRAM frames requires sending an
+HTTP/3 Settings parameter, it "sticks out". In other words, probing clients can
+learn whether a server supports HTTP Datagrams over QUIC DATAGRAM frames.
+Implementations that support this feature SHOULD always send this Settings
+parameter to avoid leaking the fact that there are applications using HTTP
+datagrams enabled on this endpoint.
 
 
 # IANA Considerations {#iana}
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-h3-datagram.md
```diff
@@ -355,11 +355,12 @@ allow endpoints to communicate their prioritization preferences.
...
-out". In other words, probing clients can learn whether a server supports this
-feature. Implementations that support this feature SHOULD always send this
-Settings parameter to avoid leaking the fact that there are applications using
-HTTP/3 datagrams enabled on this endpoint.
+Since transmitting HTTP Datagrams using QUIC DATAGRAM frames requires sending an
+HTTP/3 Settings parameter, it "sticks out". In other words, probing clients can
+learn whether a server supports HTTP Datagrams over QUIC DATAGRAM frames.
+Implementations that support this feature SHOULD always send this Settings
+parameter to avoid leaking the fact that there are applications using HTTP
+datagrams enabled on this endpoint.
```
LPardue, line 363: People elsewehere seem to keep telling me that normative requirements in security considerations are not good. So, as per my suggestion on the issue,  how about we move the recommendation up to the end of Section 5 and rephrase it to something like

> It is RECOMMENDED that implementations that support receiving HTTP Datagrams using QUIC always send the H3_DATAGRAM SETTINGS parameter with a value of 1, even if the application does not intend to use datagrams. This helps to avoid "sticking out"; see {{security}}.

Then in security just

> Since transmitting HTTP Datagrams using QUIC DATAGRAM frames requires sending an
HTTP/3 Settings parameter, it "sticks out". In other words, probing clients can
learn whether a server supports HTTP Datagrams over QUIC DATAGRAM frames.

I still don't know why sticking out is bad for security, can we add some more commentary in this section?

DavidSchinazi: Agreed, done.

LPardue (approved):
The intent is much clearer to me now, thanks!

PR #140, opened 2022-02-25
Tighten requirements on the capsule protocol

### LINKED ISSUE #128: Add to security considerations that it's not possible to generate CONNECT from JS
And point to https://fetch.spec.whatwg.org/#methods

This is an important part of the security analysis.

### LINKED ISSUE #129: Do not allow capsules for new methods
The spec currently says:

> This specification defines the "data stream" of an HTTP request as the bidirectional stream of bytes that follow the headers in both directions. In HTTP/1.x, the data stream consists of all bytes on the connection that follow the blank line that concludes either the request header section, or the 2xx (Successful) response header section. In HTTP/2 and HTTP/3, the data stream of a given HTTP request consists of all bytes sent in DATA frames with the corresponding stream ID. The concept of a data stream is particularly relevant for methods such as CONNECT where there is no HTTP message content after the headers.

In HTTP/1.x, the message content is generally delineated by the Content-Length header field or the `chunked` Transfer-Encoding; another request can follow on the same connection.  One of the (many) ways in which CONNECT and Upgrade are special is that in H1 it takes over the connection and makes the request extend to the end of the connection, preventing any further requests from being made.

The description attempts to be generic about method, but implicitly assumes an indefinite length stream of bytes which (in H1) is only true of CONNECT or Upgrade.  This concept is correct for what we have in mind, but the boundaries aren't crisply described right now.

@royfielding may have opinions.

### LINKED ISSUE #130: Capsule protocol and status codes 204,205,206
In [a PR comment](https://github.com/ietf-wg-masque/draft-ietf-masque-h3-datagram/pull/124/files#r761469804), @MikeBishop points out that status codes 204, 205, and 206 don't make sense with the capsule protocol. Should we prohibit them?

### LINKED ISSUE #131: Capsule protocol and Content-Type
In [a PR comment](https://github.com/ietf-wg-masque/draft-ietf-masque-h3-datagram/pull/124/files#r761469804), @MikeBishop points out that Content-Type doesn't make sense with the capsule protocol. Should we prohibit it?

### LINKED ISSUE #137: For HTTP/3, can datagrams be sent for existing method/upgrade tokens?
This may be answered and I just missed it. After reading #135 I think that for HTTP/1 and HTTP/2 Datagrams (via capsules) can only be sent for new methods/upgrade tokens. But it wasn't clear to me if that same property was true for HTTP/3. For example, would it be valid to send HTTP/3 datagrams along with a GET request? If so, I think this means that for HTTP/3 a GET request could have both a body AND datagrams?

DavidSchinazi:
Fixes #128.
Fixes #129.
Fixes #130.
Fixes #131.
Fixes #137.

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -27,8 +27,6 @@ author:
     organization: "Cloudflare"
     email: lucaspardue.24.7@gmail.com
 
-informative:
-  PRIORITY: I-D.ietf-httpbis-priority
 
 --- abstract
 
@@ -91,7 +89,7 @@ layer. When running over HTTP/1, requests are strictly serialized in the
 connection, therefore demultiplexing is not needed.
 
 
-# HTTP/3 DATAGRAM Format {#format}
+# HTTP/3 Datagram Format {#format}
 
 When used with HTTP/3, the Datagram Data field of QUIC DATAGRAM frames uses the
 following format (using the notation from the "Notational Conventions" section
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-h3-datagram.md
```diff
@@ -137,16 +135,23 @@ that has not yet been created, the receiver SHALL either drop that datagram
 silently or buffer it temporarily while awaiting the creation of the
 corresponding stream.
 
+HTTP/3 datagrams MUST NOT be sent associated with streams that do not define
+semantics for HTTP Datagrams. For example, existing HTTP methods GET and POST do
```
LPardue, line 139: This first sentence is a bit hard to read, maybe it's the "sent associated" term. Would this alternative also caputre your intention?

> HTTP/3 datagrams MUST NOT be sent on streams without an associated semantic for HTTP Datagrams.

draft-ietf-masque-h3-datagram.md
```diff
@@ -221,15 +225,16 @@ drop that Capsule.
 
 By virtue of the definition of the data stream, the Capsule Protocol is not in
 use on responses unless the response includes a 2xx (Successful) status code.
-The Capsule Protocol MUST NOT be used with messages that contain Content-Length
-or Transfer-Encoding header fields.
 
+The Capsule Protocol MUST NOT be used with messages that contain Content-Length,
+Content-Type, or Transfer-Encoding header fields. Additionally, HTTP status
+codes 204 (No Content), 205 (Reset Content), and 206 (Partial Content) MUST NOT
```
LPardue, line 231: Per the [HTTP style guide](https://github.com/httpwg/admin/blob/main/style-guide.md#status-codes), it is okay to omit the reason phrase in this lost i you like. I don't mind either way

```suggestion
codes 204 , 205, and 206 MUST NOT
```

draft-ietf-masque-h3-datagram.md
```diff
@@ -361,6 +367,9 @@ feature. Implementations that support this feature SHOULD always send this
 Settings parameter to avoid leaking the fact that there are applications using
 HTTP/3 datagrams enabled on this endpoint.
 
+Since use of the Capsule Protocol is restricted to new HTTP Upgrade Tokens, it is
+not accessible from JavaScript.
```
LPardue, line 371: Do we really mean all of JavaScript, or are we talking about not directly accessible rrom Web Platform APIs (Fetch, XHR etc)?

draft-ietf-masque-h3-datagram.md
DavidSchinazi: I personally find this helpful because I don't have all of the status codes memorized, so I'd be in favor of keeping the reason phrase here

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Agreed, fixed.

draft-ietf-masque-h3-datagram.md
LPardue: Works for me, kets keep them

LPardue (approved):
LGTM thanks

PR #157, opened 2022-03-24
Change argument why there is no (de)-multiplexing for HTTP/1.1

### LINKED ISSUE #156: HTTP/1.1 do not support multiplexing at all…
While the argument in Section 2 "requests are strictly serialised in the connection, therefore demultiplexing is not needed" is technically correct, Section 4. further restricts the use of HTTP/1.1 connections to a single stream of capsules/HTTP datagrams. 

I would prefer to see this restriction as an argument here.

philsbln:
Fixes #156


DavidSchinazi:
Closing as superseded by #158

PR #183, opened 2022-04-04
Be explicit about concepts in the abstract

### LINKED ISSUE #170: Abstract: Unclear use of Both
In Abstract:

"Both are intended for use by HTTP extensions, not applications."

I think instead of using Both which is unclear, rather say HTTP Datagram and Capsule Protocol. That is at least what I interpret it to be in reference to.

LPardue:
Closes #170

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -56,7 +56,8 @@ extension. When the QUIC DATAGRAM frame is unavailable or undesirable, they can
 be sent using the Capsule Protocol, a more general convention for conveying data
 in HTTP connections.
 
-Both are intended for use by HTTP extensions, not applications.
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-h3-datagram.md
```diff
@@ -56,7 +56,8 @@ extension. When the QUIC DATAGRAM frame is unavailable or undesirable, they can
 be sent using the Capsule Protocol, a more general convention for conveying data
 in HTTP connections.
 
-Both are intended for use by HTTP extensions, not applications.
+HTTP Datagrams and the Capsule protocol are intended for use by HTTP extensions,
```
DavidSchinazi, line 59:
```suggestion
HTTP Datagrams and the Capsule Protocol are intended for use by HTTP extensions,
```

DavidSchinazi (approved):

PR #186, opened 2022-04-04
Clarify no blind reencoding

### LINKED ISSUE #165: Paragraph on use of capsule protocol and http datagrams is unclear
End of section 3.5 ([The DATAGRAM Capsule](https://www.ietf.org/archive/id/draft-ietf-masque-h3-datagram-08.html#name-the-datagram-capsule)) says:

```
Note that use of the Capsule Protocol is not required to use HTTP Datagrams. 
If an HTTP extension that uses HTTP Datagrams is only defined over transports
that support QUIC DATAGRAM frames, it might not need a stream encoding.
Additionally, HTTP extensions can use HTTP Datagrams with their own data
stream protocol. However, new HTTP extensions that wish to use HTTP Datagrams
SHOULD use the Capsule Protocol unless they have a good reason not to.
```

First of all this paragraph seems to be in the wrong section because it's more generally about the capsule protocol and not the datagram capsule.

However, it's also confusing. I believe if you negotiate http datagram with the H3_DATAGRAM SETTINGS parameter, you always need to also support capsules even if QUIC datagrams are supported, as the other end (e.g. an intermediary) could send you anyway a capsule.

Or is this text meant to say that new extensions that use the http datagram but have a new settings parameter for negotiation may not need to support the capsule protocol?

I think it would be good to clarify both: what this sentence/paragraph means and what is required if you negotiate http datagrams using the H3_DATAGRAM SETTINGS parameter.

DavidSchinazi:
Fixes #165

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -401,7 +401,7 @@ it is allowed to send an HTTP Datagram and how to process them from {{format}}
 also apply to HTTP Datagrams sent and received using the DATAGRAM capsule.
 
 An intermediary can reencode HTTP Datagrams as it forwards them. In other words,
-an intermediary MAY send a DATAGRAM capsule to forward an HTTP Datagram which
+an intermediary MAY send a DATAGRAM capsule to forward an HTTP Datagram that
 was received in a QUIC DATAGRAM frame, and vice versa. Intermediaries MUST NOT
 perform this reencoding unless they have identified the use of the Capsule
 Protocol on the corresponding request stream; see {{capsule-protocol}}.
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-h3-datagram.md
```diff
@@ -402,7 +402,9 @@ also apply to HTTP Datagrams sent and received using the DATAGRAM capsule.
 
 An intermediary can reencode HTTP Datagrams as it forwards them. In other words,
 an intermediary MAY send a DATAGRAM capsule to forward an HTTP Datagram which
```
LPardue, line 404:
```suggestion
an intermediary MAY send a DATAGRAM capsule to forward an HTTP Datagram that
```

if we're touching this para :smile:

LPardue (approved):
LGTM

draft-ietf-masque-h3-datagram.md
DavidSchinazi: Gah I keep getting that one wrong!

PR #188, opened 2022-04-04
Be more honest about the mapping between app and transport

### LINKED ISSUE #172: Introduction: Missmatching description of relation to HTTP or underlying transport.
Section 1.

"When the underlying transport protocol supports unreliable delivery (such as when the QUIC DATAGRAM extension is available in HTTP/3), they can use that capability."

"This document also describes the HTTP Capsule Protocol in Section 3, to allow conveyance of HTTP Datagrams when the QUIC DATAGRAM frame is unavailable or undesirable, such as when earlier versions of HTTP are in use."

To me there is a bit of mismatch between these two sentences and what HTTP/3 is prior to this extension.

The first indicate that datagram support is a capability of the underlying transport for the HTTP version but then link it hard to HTTP/3. While the HTTP/3 support is dependent on this extension in addition to a QUIC stack having Datagram support. While the second takes it as a property of the HTTP version, where it is assumed that earlier version can't support this, i.e. those that are defined to use TCP only. 

Technically a HTTP/3 endpoint with a QUIC stack without datagram support can  use capsules and this document does not require an endpoint to support QUIC datagram.

And the second statement should likely use a statement saying like …  unavailable or undesirable, such as when TCP is used as transport protocol as with HTTP/1.x and HTTP/2.

LPardue:
Closes #172

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -76,7 +76,7 @@ are not part of message content; instead, they are intended for use by HTTP
 extensions (such as the CONNECT method), and are compatible with all versions of
 HTTP.
 
-When HTTP is based on top of a transport protocol that supports unreliable
+When HTTP is running over a transport protocol that supports unreliable
 delivery (such as when the QUIC DATAGRAM extension is available to HTTP/3), HTTP
 Datagrams can use that capability.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-h3-datagram.md
```diff
@@ -74,13 +74,17 @@ supports the bidirectional and possibly multiplexed exchange of data inside an
 HTTP connection. While HTTP datagrams are associated with HTTP requests, they
 are not part of message content; instead, they are intended for use by HTTP
 extensions (such as the CONNECT method), and are compatible with all versions of
-HTTP. When the underlying transport protocol supports unreliable delivery (such
-as when the QUIC DATAGRAM extension is available in HTTP/3), they can use that
-capability.
+HTTP.
+
+When HTTP is based on top of a transport protocol that supports unreliable
```
DavidSchinazi, line 79:
```suggestion
When HTTP is running over a transport protocol that supports unreliable
```

DavidSchinazi (approved):
Looks great modulo one tweak

PR #191, opened 2022-04-04
Switch to QUIC IANA policy

### LINKED ISSUE #189: IANA Policy for new Capsule Type registry
When the first IANA registry was added to this document, it used "First Come First Served" to save time. This got carried over through the various changes and now the new "Capsule Type" registry is still marked that way. But in that corresponding time, the QUIC WG came up with a very good IANA policy for varint registries and documented it in RFC 9000. We should adopt that policy for Capsule Types.

DavidSchinazi:
Closes #189


LPardue (approved):

DavidSchinazi:
Seeing support and hearing no pushback, merging this as today marks the end of WGLC

PR #193, opened 2022-04-05
Clarify that the optionality of capsules applies to extensions

### LINKED ISSUE #167: SHOULD or MUST intermediaries forward unknown capsule types without modification?
The first SHOULD should be a MUST because there is a clear exception specified but in all other cases this is a MUST.

The second SHOULD should really be a MUST. Or why is that a SHOULD?
If there is a reason for a SHOULD this needs to be explained but I think it would break interoperability.

DavidSchinazi:
Fixes #167


LPardue:
This is a good step but I think it overlooks one of the points that Magnus articulated. If an HTTP extension does what we say here, it might fall foul of intermediaries and protocol version conversion. That not some *we* have to solve but I suspect if we can concisely describe those considerations we would address the issue.

DavidSchinazi:
Good point, added text

LPardue (approved):

gloinul:
Thanks, this is much clearer and is a descent explanation of the issues and motivation behind the should. I guess a more correct part will require much more text.

PR #194, opened 2022-04-05
Clarify that there is no content with capsules

### LINKED ISSUE #169: HTTP content handling capability with Capsule protocol
I find the difference between HTTP 1.x and HTP/2 vs HTTP/3 when an endpoint support QUIC datagram quite worrying. So as capsules are sent instead of the message body in requests and response, for HTTP/3 when one have QUIC datagram support a method can be defined to use both the reliable body as well as HTTP datagrams. This is not possible over HTTP/3 when the QUIC stack doesn't support QUIC datagram or for HTTP 1.x or HTTP/2. I think this makes it very hard to define something that will just work. I Think the easy fix for this is to just define in addition to the Datagram capsule a body capsule that carries a fragment of the body, so that one can send over an upgraded path both parts of the body as well as the HTTP Datagrams. The fact that the Capsule protocol doesn't give feature parity even for HTTP/3 is deeply worrying in my view. 

I can understand how this has not come up prior as the existing HTTP methods that been targeted by this mechanism, i.e. Connect results in that the data stream becomes the capsule streams. However, if some other HTTP extension that using Extended Connect are used, then the need for both body and HTTP datagram could arise.

DavidSchinazi:
Fixes #169

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -336,7 +336,7 @@ drop that Capsule and skip over it to parse the next Capsule.
 
 By virtue of the definition of the data stream:
 
-* The Capsule Protocol is not in use on responses unless the response includes a
+* The Capsule Protocol is not in use unless the response includes a
   2xx (Successful) status code.
 
 * When the Capsule Protocol is in use, the associated HTTP request and response
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -336,8 +336,8 @@ drop that Capsule and skip over it to parse the next Capsule.
 
 By virtue of the definition of the data stream:
 
-* The Capsule Protocol is not in use unless the response includes a
-  2xx (Successful) status code.
+* The Capsule Protocol is not in use unless the response includes a 2xx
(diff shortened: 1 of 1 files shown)


gloinul:
This looks ok

draft-ietf-masque-h3-datagram.md
```diff
@@ -334,8 +334,14 @@ forward Capsules with an unknown Capsule Type without modification.
 Endpoints which receive a Capsule with an unknown Capsule Type MUST silently
 drop that Capsule and skip over it to parse the next Capsule.
 
-By virtue of the definition of the data stream, the Capsule Protocol is not in
-use on responses unless the response includes a 2xx (Successful) status code.
+By virtue of the definition of the data stream:
+
+* The Capsule Protocol is not in use on responses unless the response includes a
+  2xx (Successful) status code.
```
LPardue, line 340: Strictly, the capsule protocol isn't in use at all unless the response is 2xx right?

Maybe just simplify this statement further.
```suggestion
* The Capsule Protocol is not in use unless the response includes a
  2xx (Successful) status code.
```

LPardue (approved):
LGTM

PR #197, opened 2022-05-25
reword sentence

martinduke:
Nit from AD review:

The original sentence is ambiguously worded, and could be interpreted as saying that there could not be normal HTTP requests prior to the capsule protocol.


LPardue (approved):
Thanks, this is an important clarification that makes clearer the original intent of the statement.

DavidSchinazi (approved):

PR #203, opened 2022-05-31
Add reference that explains what HTTP extensions are

### LINKED ISSUE #202: Add reference for HTTP Extensions
Issue created in reponse to [David Mandelberg's SECDIR review](https://mailarchive.ietf.org/arch/msg/masque/ohX2BAveuPxj9CiZaRjf92G6ma4/)

> Reviewer: David Mandelberg
> Review result: Ready
>
> I did not notice any security issues with this document, and the security
considerations section seems straightforward.
>
> (nit / question / feel free to ignore) I was a bit confused by the term "HTTP
extensions" at first, but from context, I'm guessing it's used loosely to mean
any addition to HTTP? I found RFC 2774 when I was looking to see if it meant
something specific, but I'm guessing it's not a reference to a 20+ year old
experimental RFC?

The term HTTP extensions is defined in [Section 16 of HTTP](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-semantics-19#section-16). Since we already have a normative reference to that document, I propose to resolve this issue by adding a reference to that section in our introduction.

DavidSchinazi:
Closes #202

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -73,11 +73,12 @@ not applications.
 
 # Introduction {#intro}
 
-HTTP extensions sometimes need to access underlying transport protocol features
-such as unreliable delivery (as offered by {{!DGRAM=RFC9221}}) to enable
-desirable features. For example, this could allow introducing an unreliable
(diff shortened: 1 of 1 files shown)


LPardue (approved):
Lgtm

PR #205, opened 2022-06-03
Be clearer about message error vs. stream errors

### LINKED ISSUE #200: Error Handling section is sloppy with terms
In Sec 3.3:

When an error occurs in processing the Capsule Protocol, the receiver
   MUST treat the message as malformed or incomplete...

there are 2 other reference to malformed "messages" as a result of Capsule shenanigans.

but in the introduction it says that datagrams "are not part of message content". 

So what is "the message" that should be treated this way? I think you mean the Request or Response that initiated the Capsule protocol.

If that's correct, a more precise phrasing might be "MUST act as if the message that initiated the capsule protocol was malformed".

LPardue:
Explicitly replicates parts of H2 and H3 CONNECT stream definition to
make it clear what frame types are/are not allowed on a stream.

Mostly anything else can't happen but if it does, approach error
handling _as if it were like_ a malformed message (I.e whether to
terminate the stream or whatever).

Fixes #200

### DIFF
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -373,7 +373,7 @@ stream management frames (RST_STREAM, WINDOW_UPDATE, and PRIORITY) are
 permitted; other frame types MUST NOT be sent and MUST be treated as a stream
 error ({{Section 5.4.2 of H2}}) if received. In HTTP/3, only DATA frames are
 permitted; other frame types MUST NOT be sent and MUST be treated as connection
-error o)f type H3_FRAME_UNEXPECTED ({{Section 8 of H3}}) if received. In both
+error of type H3_FRAME_UNEXPECTED ({{Section 8 of H3}}) if received. In both
 cases, extension frame types MAY be used if a specifically permitted by the
 definition of the extension.
 
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -374,7 +374,7 @@ permitted; other frame types MUST NOT be sent and MUST be treated as a stream
 error ({{Section 5.4.2 of H2}}) if received. In HTTP/3, only DATA frames are
 permitted; other frame types MUST NOT be sent and MUST be treated as connection
 error of type H3_FRAME_UNEXPECTED ({{Section 8 of H3}}) if received. In both
-cases, extension frame types MAY be used if a specifically permitted by the
+cases, extension frame types MAY be used if specifically permitted by the
 definition of the extension.
 
 Each capsule's payload MUST contain exactly the fields identified in its
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -367,7 +367,7 @@ in {{Section 4.1.3 of H3}}. For HTTP/2, the handling of malformed messages is
 described in {{Section 8.1.1 of H2}}. For HTTP/1.1, the handling of incomplete
 messages is described in {{Section 8 of H1}}.
 
-For HTTP/2 and HTTP/3, once a both endpoints agree to use the Capsule Protocol,
+For HTTP/2 and HTTP/3, once both endpoints agree to use the Capsule Protocol,
 the frame usage requirements of the stream change. In HTTP/2, only DATA or
 stream management frames (RST_STREAM, WINDOW_UPDATE, and PRIORITY) are
 permitted; other frame types MUST NOT be sent and MUST be treated as a stream
--- a/draft-ietf-masque-h3-datagram.md
+++ b/draft-ietf-masque-h3-datagram.md
@@ -359,23 +359,18 @@ violation of these requirements MUST treat the HTTP message as malformed.
 
 ## Error Handling
 
-Unless otherwise stated, Capsule Protocol endpoints are expected to apply the
-same message format rules as the underlying HTTP version. When a receiver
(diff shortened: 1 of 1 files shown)


draft-ietf-masque-h3-datagram.md
```diff
@@ -359,13 +359,24 @@ violation of these requirements MUST treat the HTTP message as malformed.
...
 described in {{Section 8.1.1 of H2}}. For HTTP/1.1, the handling of incomplete
 messages is described in {{Section 8 of H1}}.
 
+For HTTP/2 and HTTP/3, once a both endpoints agree to use the Capsule Protocol,
+the frame usage requirements of the stream change. In HTTP/2, only DATA or
+stream management frames (RST_STREAM, WINDOW_UPDATE, and PRIORITY) are
+permitted; other frame types MUST NOT be sent and MUST be treated as a stream
+error ({{Section 5.4.2 of H2}}) if received. In HTTP/3, only DATA frames are
+permitted; other frame types MUST NOT be sent and MUST be treated as connection
+error o)f type H3_FRAME_UNEXPECTED ({{Section 8 of H3}}) if received. In both
```
tfpauly, line 376:
```suggestion
error of type H3_FRAME_UNEXPECTED ({{Section 8 of H3}}) if received. In both
```

draft-ietf-masque-h3-datagram.md
```diff
@@ -359,13 +359,24 @@ violation of these requirements MUST treat the HTTP message as malformed.
...
-{{Section 4.1.3 of H3}}. For HTTP/2, the handling of malformed messages is
+Unless otherwise stated, Capsule Protocol endpoints are expected to apply the
+same message format rules as the underlying HTTP version. When a receiver
+encounters a processing error, it MUST be treated as if it were a malformed or
+incomplete message. For HTTP/3, the handling of malformed messages is described
+in {{Section 4.1.3 of H3}}. For HTTP/2, the handling of malformed messages is
 described in {{Section 8.1.1 of H2}}. For HTTP/1.1, the handling of incomplete
 messages is described in {{Section 8 of H1}}.
 
+For HTTP/2 and HTTP/3, once a both endpoints agree to use the Capsule Protocol,
```
martinduke, line 370:
```suggestion
For HTTP/2 and HTTP/3, once both endpoints agree to use the Capsule Protocol,
```

martinduke (changes requested):
I'm having github problems but there's another s/MUST be treated as/MSUT be treated as if it were

martinduke (approved):

draft-ietf-masque-h3-datagram.md
```diff
@@ -359,23 +359,35 @@ violation of these requirements MUST treat the HTTP message as malformed.
...
 ## Error Handling
 
-When an error occurs in processing the Capsule Protocol, the receiver MUST treat
-the message as malformed or incomplete, according to the underlying transport
-protocol. For HTTP/3, the handling of malformed messages is described in
-{{Section 4.1.3 of H3}}. For HTTP/2, the handling of malformed messages is
+Unless otherwise stated, Capsule Protocol endpoints are expected to apply the
+same message format rules as the underlying HTTP version. When a receiver
+encounters a processing error, it MUST be treated as if it were a malformed or
+incomplete message. For HTTP/3, the handling of malformed messages is described
```
DavidSchinazi, line 365: I'm not sure this first sentence adds much value, because it sounds to me like "please follow referenced specs unless told not to"

Also I'd specifically call out the capsule protocol in the normative sentence that follows

```suggestion
When a receiver encounters an error processing the Capsule Protocol, the receiver
MUST treat it as if it had received a malformed or incomplete HTTP message.
For HTTP/3, the handling of malformed messages is described
```

draft-ietf-masque-h3-datagram.md
```diff
@@ -359,23 +359,35 @@ violation of these requirements MUST treat the HTTP message as malformed.
...
 
+For HTTP/2 and HTTP/3, once both endpoints agree to use the Capsule Protocol,
+the frame usage requirements of the stream change. In HTTP/2, only DATA or
+stream management frames (RST_STREAM, WINDOW_UPDATE, and PRIORITY) are
+permitted; other frame types MUST NOT be sent and MUST be treated as a stream
+error ({{Section 5.4.2 of H2}}) if received. In HTTP/3, only DATA frames are
+permitted; other frame types MUST NOT be sent and MUST be treated as connection
+error of type H3_FRAME_UNEXPECTED ({{Section 8 of H3}}) if received. In both
+cases, extension frame types MAY be used if specifically permitted by the
+definition of the extension.
```
DavidSchinazi, line 378: This text duplicates normative text from our references, how about the below? Also, I'd move it to the end of `{{capsule-protocol}}` since it isn't specific to error handling.

```suggestion
Since the Capsule Protocol only applies to definitions of new HTTP Upgrade
Tokens, in HTTP/2 and HTTP/3 it can only be used with the CONNECT method.
Therefore, once both endpoints agree to use the Capsule Protocol, the frame
usage requirements of the stream change as specified in {{Section 8.5 of H2}}
and {{Section 4.2 of H3}}.
```

LPardue: Oh yes, that is much better

draft-ietf-masque-h3-datagram.md
LPardue: With the change below, this suggestion is better, thanks.

